PR #26, opened 2021-05-04
BUGFIX: compile error in the Asian version of MSVC

BenjaminGu:


RandyGaul:
Looks like a safe thing to merge in, thanks!

PR #71, opened 2022-11-02
Fix invalid c99 syntax

jonstvns:
This `enum my_type : int` is not valid syntax in C


RandyGaul:
Ah thank you! I was experimenting and forgot to remove this

PR #72, opened 2022-11-06
Add cimgui support

jonstvns:
- add cimgui
- replace usage of `libraries/imgui` with `libraries/cimgui/imgui`
- move `imgui_impl_sdl.h/cpp` to `libraries/cimgui/imgui/backends
- add `CUTE_FRAMEWORK_CIMGUI` option to cmake to build cute framework with the cimgui implementation (default OFF)
- fix ordering of shutdown between sokol imgui and sdl2 impl imgui


RandyGaul:
This is really cool, thanks for all the effort! Love the addition.

PR #119, opened 2023-06-25
Fix shared lib builds

jonstvns:
The implementation define for cute_net.h was wrong which somehow caused linker errors on sharedlib builds but not static


RandyGaul:
Thank you!

PR #128, opened 2023-08-15
Fix cf_hashtable_has macro

empyreanx:
Previously the macro `cf_hashtable_has` was set to use the `cf_hashtable_remove_impl`function


RandyGaul:
Thank you looks great! Though hfree maybe needs to be called at the end of the test?

empyreanx:
Good catch, I'll make that change right now

empyreanx:
Should be good to go

PR #131, opened 2023-09-09
cf_path_get_filename - support path string arguments with no "/" seperators

jonstvns:
If a user passes in a path string to `cf_path_get_filename` that contains no "/" separators, assume that string is the filename. Return a copy of that string instead of `NULL`

### DIFF
--- a/src/cute_file_system.cpp
+++ b/src/cute_file_system.cpp
@@ -34,8 +34,9 @@
 
 char* cf_path_get_filename(const char* path)
 {
+	if (!path || path[0] == '\0') { return NULL; }
+
 	int at = slast_index_of(path, '/');
-	if (at == -1) return NULL;
 	return smake(path + at + 1);
 }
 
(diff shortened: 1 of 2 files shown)


jonstvns:
hmm, this fails [this test](https://github.com/RandyGaul/cute_framework/blob/master/test/test_path.cpp#L42C6-L42C13) because `.txt` is assumed to not be a filename, but I don't think that's true. For example,  `.gitignore` is a valid filename

jonstvns:
It also fails [this test](https://github.com/RandyGaul/cute_framework/blob/master/test/test_path.cpp#L38) because it assumes `file.txt` is not a filename

RandyGaul:
Does it work with the virtual file system as well? I remember paths start with a slash, but maybe it supports this too. Please confirm and also update the tests.

jonstvns:
This does work with the VFS. If you mount something at "", you can access a top level file by just passing "file.txt". No leading "/" required

PR #136, opened 2023-11-05
Return the same camera dimensions that were passed in

waldnercharles:
`cf_camera_dimensions` appears to divide the passed in width and height by 2. This PR just makes it so that when we peek at the dimensions, we get the same thing we passed in.

```c++
cf_camera_dimensions(320, 180);
v2 dim = cf_camera_peek_dimensions();
ImGui::Text("%f, %f", dim.x, dim.y); // displays 160, 90
```


waldnercharles:
I have a test written for this as well but, it does require a GFX context. If you'd like me to include it, let me know.

PR #140, opened 2023-11-11
Add an AABB grid

waldnercharles:
This is my first stab at an aabb grid I made while using CF. Please give critical feedback. Feelings won't get hurt. 😅

### DIFF
--- a/include/cute_aabb_grid.h
+++ b/include/cute_aabb_grid.h
@@ -25,44 +25,50 @@
 #include <cute_math.h>
 #include <cute_array.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+typedef struct CF_AabbGrid { uint64_t  id; } CF_AabbGrid;
+typedef struct CF_AabbGridNode { int id; } CF_AabbGridNode;
+typedef bool (CF_AabbGridQueryFn)(CF_AabbGridNode node, CF_Aabb aabb, void* leaf_udata, void* fn_udata);
+
+CF_API CF_AabbGrid CF_CALL cf_make_aabb_grid(CF_V2 pos, int width, int height, int grid_size);
+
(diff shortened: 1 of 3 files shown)


src/cute_aabb_grid.cpp
```diff
@@ -0,0 +1,194 @@
...
+	for (int x = min_x; x <= max_x; x++)
+	{
+		for (int y = min_y; y <= max_y; y++)
+		{
+			int index = y * max_tiles_x + x;
+
+			for (auto b : grid->map[index])
+			{
+				bool already_visited = false;
+				for (auto v : visited)
```
waldnercharles, line 162: I'd normally use `std::find(...)` here but, I actually have a `cf_algorithms.h` in my project to do the iteration myself so I don't have to worry about all the craziness that `std::find` is doing.

include/cute_aabb_grid.h
```diff
@@ -0,0 +1,77 @@
...
+typedef struct CF_AabbGridNode { int id; } CF_AabbGridNode;
+typedef bool (CF_AabbGridQueryFn)(CF_AabbGridNode node, CF_Aabb aabb, void* leaf_udata, void* fn_udata);
+
+CF_API CF_AabbGrid CF_CALL cf_make_aabb_grid(CF_V2 pos, int width, int height, int grid_size);
+
+CF_API void CF_CALL cf_destroy_aabb_grid(CF_AabbGrid grid_handle);
+
+CF_API void CF_CALL cf_aabb_grid_clear(CF_AabbGrid grid);
+
+CF_API void CF_CALL cf_aabb_grid_set_pos(CF_AabbGrid grid_handle, CF_V2 pos);
```
waldnercharles, line 42: It's not very obvious but, this function invalidates the `aabb_grid`. I couldn't think of an API/implementation that I liked, so I just settled on the simplest for now. I was tempted to have something like `cf_aabb_grid_update` or `cf_aabb_grid_clear_and_move` to make it clear that moving the grid invalidates the grid.

PR #144, opened 2023-11-19
Remove ssize

### LINKED ISSUE #143: Building with C++ fails
I'm trying to build my project with C++20 with cute in there, and, I noticed that `cute_string.h` does a `#define ssize(s) cf_string_size(s)`. It looks like C++20 defines its own `ssize` in `xutility`
```c++
#if _HAS_CXX20
_EXPORT_STD template <class _Container>
_NODISCARD constexpr auto ssize(const _Container& _Cont) noexcept(noexcept(
    static_cast<common_type_t<ptrdiff_t, make_signed_t<decltype(_Cont.size())>>>(_Cont.size()))) /* strengthened */
    -> common_type_t<ptrdiff_t, make_signed_t<decltype(_Cont.size())>> {
    using _Common = common_type_t<ptrdiff_t, make_signed_t<decltype(_Cont.size())>>;
    return static_cast<_Common>(_Cont.size());
}

_EXPORT_STD template <class _Ty, ptrdiff_t _Size>
_NODISCARD constexpr ptrdiff_t ssize(const _Ty (&)[_Size]) noexcept {
    return _Size;
}
#endif // _HAS_CXX20
```

The gist is, `#define ssize(s)` is overriding the `ssize` in `xutility`

waldnercharles:
This is for C++20 support. I was going to add something to the build.yml to test builds with C++20, but wasn't sure what the best way to do that would be. I was tempted to just create 4 more actions with `-DCMAKE_CXX_STANDARD=20` specified, but that felt like overkill.

fixes #143


waldnercharles:
It appears there may be some more ambiguous calls that aren't exposed by the examples and tests. For instance, `cute_array.h` seems to have an ambiguous reference to `move()`

PR #149, opened 2024-01-07
Cast data to uintptr_t before doing arithemtic

### LINKED ISSUE #138: Clang build fails
I was just setting up github-actions for the first time ever (so exciting) when I noticed that CF doesn't build successfully on clang.
```
build/_deps/cute-src/libraries/cute/cute_tls.h:1248:43: error: arithmetic on a pointer to void
        int w = s2n_send(ctx->connection, data + bytes_written, size - bytes_written, &blocked);
```

The workflow for CF uses gcc, which doesn't complain; or treats that error as a warning instead. It's not a priority for me to fix since I don't actually care about clang, but, it was one of the default compilers that github setup for me.

bullno1:
Fixes #138


RandyGaul:
Great!

PR #150, opened 2024-01-07
Ensure that font is always pushed in scratch example

bullno1:
Since the stacks of states always get reset every frame, the scratch example will crash if it only push the font once.


RandyGaul:
Thanks!

PR #159, opened 2024-06-29
WIP

ogam:
Slight modifications to cf_draw for console and input texts.

This allows incomplete xml tags to not cause crashes during sanitations.
Able to turn on and off text effects to see what xml tags you are currently typing out.
s_draw_text now returns last glyph cursor position and uses text_length when drawing glyphs.
s_parse_code_val has extra string checks incase user is still typing out xml tags and string is empty, along with default color hex and double values to be set to 0.


ogam:
i haven't touched this in a while since doing some UI some months ago, thinking about this again might be better to use xml tags for the selection box text. using the above you have to call cf_text_width() twice to grab the start and end position to draw a box.

RandyGaul:
This looks pretty understandable and I think I can merge it in as-is. Could you elaborate a bit more on what you meant by calling cf_text_width twice? I'm not sure I understand this point

PR #171, opened 2024-07-22
fix: implied return and NULL cast to uint64_t

alexlnkp:
- The wrapper functions in `cute_draw.h` are meant to return bool type, however the return values of internal `cf_*` functions which are called aren't stored or returned, thus making these functions return nothing. Easily fixed by calling the internal functions within the wrapper and returning the returned values.

- In `cute_audio.cpp`, a NULL was returned previously. Due to `CF_Audio` having just one field in the struct for the data, that would mean the NULL would be cast to uint64_t. This means that comparing the return value from `cf_audio_load_*` functions to NULL would not yield any actual results in some scenarios, as the value would be cast to `uint64_t`, resulting in undefined behavior. Solution is simple, just return an empty `CF_Audio` struct. Then, if needed, the return value can simply be compared to 0, which would imply that either the `CF_Audio` wasn't yet populated, OR there was an error trying to load audio. (NOTE: Asserts can also be used here to indicate what is PRECISELY the case.)


RandyGaul:
Thank you! These are helpful adjustments.

PR #178, opened 2024-08-01
Use correct directory when copying sample data

thealpaca89:
This commit updates the CMakeLists.txt file to use the correct directory when copying over the sample data when building from source directory.

Previously, the sample data was copied from the top-level CMake working dir, which wouldn't work when the samples were being built from a project including CF.

With this change, the sample data should be copied over from the CF working directory, which should work in all cases.


RandyGaul:
This bug has been annoying for quite some time, thank you very much for helping with it!

PR #179, opened 2024-08-04
Modular cmakelists

pusewicz:
@RandyGaul Hey Randy, inspired by [this post about CMake](https://www.foonathan.net/2022/06/cmake-fetchcontent/), I went ahead and restructured the CMake configuration. The main takeaways are:

- Configuration is now split per directory of interest (src, test, samples)
- No build of test and samples when Cute is used as a dependency

The upside here is that the main CMakeLists.txt is now smaller and easier to parse (although I still think it could become even simpler) and that Cute would not even attempt to build the tests and samples unless needed.

The downside is that, well, the config is now split across multiple files, so it's necessary to know which one to edit to make changes.

Please let me know if this is a change you'd like merged, any changes you'd like to see made or close it if this is not something you see useful. Thanks!

Remaining todo:

- [ ] Fix CI builds

### DIFF
--- a/include/cute_audio.h
+++ b/include/cute_audio.h
@@ -521,9 +521,9 @@ CF_INLINE void audio_set_pause(bool true_for_paused) { cf_audio_set_pause(true_f
 // -------------------------------------------------------------------------------------------------
 
 CF_INLINE void music_play(Audio audio_source, float fade_in_time = 0) { cf_music_play(audio_source, fade_in_time); }
-CF_INLINE void music_stop(float fade_out_time = 0) { cf_music_stop(fade_out_time = 0); }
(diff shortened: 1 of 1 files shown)


RandyGaul:
I do feel like a single cmake file is a little easier to maintain and slightly simpler. Is it possible to keep any improvements you've made while also keeping a single cmake file? If so, could you re-summarize what the benefits are after refactoring back to a single cmake file?

Thanks for taking the time to think about how to improve CF!

pusewicz:
@RandyGaul It absolutely is possible to get those benefits. I'll work on that and rebase the branch.

One thing that's currently annoying is that the failing commands on Windows builds do not mark the build as... failed. Not sure why that is.

PR #182, opened 2024-08-05
Bump SDL2 to 2.30.5

pusewicz:
What it says on the 📦.


RandyGaul:
Looks like it doesn’t quite succeed for MSVC

RandyGaul:
I triggered a rerun just to see if it was sporadic

pusewicz:
@RandyGaul Looks like a regression that got fixed https://github.com/libsdl-org/SDL/issues/10462. We'll have to wait for the next release.

PR #183, opened 2024-08-07
Use bash on Windows

pusewicz:
This is so that the scripts can fail if commands return non-zero.

Cmd happily continues and never returns failures.

Ref https://github.com/actions/runner-images/issues/6668

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,7 @@ jobs:
       
       # Run tests for MSVC 17
       - name: Run Tests for MSVC 17
-        shell: cmd
+        shell: bash
         run: |
           cd build_folder\Release
           tests.exe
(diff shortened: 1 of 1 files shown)


RandyGaul:
Bash isn't available on windows by default, so I'm not sure if this is a good solution. Are you sure it can't return failures? I've noticed the script runner here on github fail for windows a variety of times. Going to close this for now since I don't think we can swap the CI to bash here. Please re-open this if I'm mistaken.

pusewicz:
Yeah, I need to open the PR to get the build started, so that was my only way to test this.

pusewicz:
@RandyGaul

> Bash isn't available on windows by default, so I'm not sure if this is a good solution

It is not available by default, but it is on the GitHub images. As you can see in the build failure, it's trying to cd into a directory and failing. The cd commands would need tweaking to resemble Unix paths I believe.

PR #192, opened 2024-08-22
Resize vertex buffer if necessary in cf_mesh_update_vertex_data

waldnercharles:
I don't know if this is the right way to do this. It solved my problem so I figured I'd open a PR to discuss. Would it be better to instead do something like a pow2 resize?

### DIFF
--- a/src/cute_graphics.cpp
+++ b/src/cute_graphics.cpp
@@ -1312,12 +1312,32 @@ void cf_mesh_set_attributes(CF_Mesh mesh_handle, const CF_VertexAttribute* attri
 
 void cf_mesh_update_vertex_data(CF_Mesh mesh_handle, void* data, int count)
 {
-	// @TODO Destroy/recreate for a larger size.
-
 	// Copy vertices over to the driver.
 	CF_MeshInternal* mesh = (CF_MeshInternal*)mesh_handle.id;
 	CF_ASSERT(mesh->attribute_count);
 	int size = count * mesh->vertices.stride;
+
+	// Resize buffers if necessary
+	if (size > mesh->vertices.size)
+	{
+		SDL_GpuReleaseBuffer(app->device, mesh->vertices.buffer);
+		SDL_GpuReleaseTransferBuffer(app->device, mesh->vertices.transfer_buffer);
+
+		mesh->vertices.size = size;
+		SDL_GpuBufferCreateInfo buf_info = {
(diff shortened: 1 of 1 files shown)


RandyGaul:
Looks good, though I would agree yes, a 2x size would be a bit better. Go ahead and just double the size if this if-statement is hit. The default size is very small like 5mb, so I would expect for games with lots of sprites for this to get hit a few times before settling.

src/cute_graphics.cpp
```diff
@@ -1312,12 +1312,32 @@ void cf_mesh_set_attributes(CF_Mesh mesh_handle, const CF_VertexAttribute* attri
...
-	// @TODO Destroy/recreate for a larger size.
-
 	// Copy vertices over to the driver.
 	CF_MeshInternal* mesh = (CF_MeshInternal*)mesh_handle.id;
 	CF_ASSERT(mesh->attribute_count);
 	int size = count * mesh->vertices.stride;
+
+	// Resize buffers if necessary
+	if (size > mesh->vertices.size)
+	{
```
RandyGaul, line 1322: Nit: Brace on same line as predicate

src/cute_graphics.cpp
```diff
@@ -1312,12 +1312,32 @@ void cf_mesh_set_attributes(CF_Mesh mesh_handle, const CF_VertexAttribute* attri
 
 void cf_mesh_update_vertex_data(CF_Mesh mesh_handle, void* data, int count)
 {
-	// @TODO Destroy/recreate for a larger size.
-
 	// Copy vertices over to the driver.
 	CF_MeshInternal* mesh = (CF_MeshInternal*)mesh_handle.id;
 	CF_ASSERT(mesh->attribute_count);
 	int size = count * mesh->vertices.stride;
```
RandyGaul, line 1318: Let's just add `* 2` to the end of this here

src/cute_graphics.cpp
pusewicz: Speaking of this, @RandyGaul how do you feel about adding linters and formatters so the code is always consistent?

RandyGaul: Ehh not the biggest fan since it takes a lot of time to setup and maintain

pusewicz: But then it makes life easier, and with github actions it can be validates with each PR.

PR #204, opened 2024-08-30
Update to the renamed GPU functions and pull directly from libsdl-org

waldnercharles:


src/cute_app.cpp
```diff
@@ -254,7 +254,7 @@ CF_Result cf_make_app(const char* window_title, CF_DisplayID display_id, int x,
 		} else if (use_vulkan) {
 			device_name = "Vulkan";
 		}
-		device = SDL_GpuCreateDevice(SDL_ShaderCross_GetShaderFormats(), true, false, device_name);
+		device = SDL_CreateGPUDevice(SDL_ShaderCross_GetShaderFormats(), false, device_name);
```
waldnercharles, line 257: This function loses a boolean, but, I couldn't find what the boolean was doing previously so I just omit it. The samples seemed to run fine without it.

libraries/cute/cute_sync.h
```diff
@@ -532,17 +536,20 @@ cute_cv_t cute_cv_create()
 
 int cute_cv_wake_all(cute_cv_t* cv)
 {
-	return !SDL_BroadcastCondition((SDL_Condition*)cv->align);
+	SDL_BroadcastCondition((SDL_Condition*)cv->align);
+	return 1;
```
waldnercharles, line 540: Just wanted to call out that this API changed as well.

src/cute_app.cpp
```diff
@@ -233,11 +233,11 @@ CF_Result cf_make_app(const char* window_title, CF_DisplayID display_id, int x,
 	// Turn on high DPI support for all platforms.
 	options |= SDL_WINDOW_HIGH_PIXEL_DENSITY;
 	
-	if (SDL_Init(sdl_options)) {
+	if (!SDL_Init(sdl_options)) {
```
waldnercharles, line 236: SDL_Init and SDL_InitSubsystem (Along with some of the other wrappers around InitSubsystem) seem to return true on success now.

PR #207, opened 2024-09-02
Check support for texture format

pusewicz:
Similarly to https://github.com/cocos2d/cocos2d-x/pull/19774/files, we default to `CF_PIXEL_FORMAT_D32_FLOAT_S8_UINT` and set it to `CF_PIXEL_FORMAT_D24_UNORM_S8_UINT` if it's supported.

### DIFF
--- a/src/cute_graphics.cpp
+++ b/src/cute_graphics.cpp
@@ -509,6 +509,16 @@ CF_BackendType cf_query_backend()
 	}
 }
 
+bool cf_texture_supports_format(CF_PixelFormat format, CF_TextureUsageBits usage)
+{
+	return SDL_GPUTextureSupportsFormat(
+			app->device,
+			s_wrap(format),
+			SDL_GPU_TEXTURETYPE_2D,
+			usage
+			);
+}
+
 CF_TextureParams cf_texture_defaults(int w, int h)
 {
 	CF_TextureParams params;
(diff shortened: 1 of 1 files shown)


pusewicz:
This needs some more work:

```

-[MTLDebugRenderCommandEncoder setRenderPipelineState:]:1615: failed assertion `Set Render Pipeline State Validation
For stencil attachment, the render pipeline's pixelFormat (MTLPixelFormatInvalid) does not match the framebuffer's pixelFormat (MTLPixelFormatDepth32Float_Stencil8).
```

RandyGaul:
Looks good so far. Not sure why the depth format of a pipeline doesn't match. The depth format is supposed to get copied over here: https://github.com/RandyGaul/cute_framework/blob/master/src/cute_graphics.cpp#L1651

pusewicz:
It raises here: https://github.com/RandyGaul/cute_framework/blob/master/src/cute_graphics.cpp#L1793

RandyGaul:
Yep so it's a matter of debugging to see where the different depth formats are set. If you need help debugging let me know, or if you're stuck I can also look into it and solve it. Thanks so much for the help!

pusewicz:
@RandyGaul It turns out, this branch works just fine. I applied the same branch again, on a fresh clone of Cute and it works as expected.
Screenshot 2024-09-02 at 19 42 32

PR #212, opened 2024-09-03
Resize the transfer buffer in s_update_buffer

waldnercharles:
This one might need a closer look. It felt a little strange to me to just resize the transfer buffer without checking its size, but, there was no size on the mesh object for it. I'm assuming we'll add a size and then only resize the transfer buffer when any of the 3 buffer sizes exceed the transfer buffer size, but, the index and instance functions still seem to be in flux so I just figured I'd open this PR to get the conversation started.

### DIFF
--- a/src/cute_graphics.cpp
+++ b/src/cute_graphics.cpp
@@ -1315,7 +1315,7 @@ void cf_destroy_mesh(CF_Mesh mesh_handle)
 	CF_FREE(mesh);
 }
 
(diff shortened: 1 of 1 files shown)


waldnercharles:
This PR will likely fail to build until one of the other PRs is merged that fixes the SDL renames.

RandyGaul:
I'd prefer to just revert to 3 transfer buffers as mentioned in discord. Changing to a single buffer was an oversight on my part. Going to close this and reimplement it now.

RandyGaul:
https://github.com/RandyGaul/cute_framework/commit/41bf5a999048d253d883af5c6a9f4aed911b38e2

PR #215, opened 2024-09-06
Dynamic stencil reference and blend constants

pusewicz:
Addresses changes in https://github.com/libsdl-org/SDL/pull/10704

This change adds two new methods to dynamically adjust the reference and blend constants as described in the PR linked above.

I admit this is beyond my current non-existent understanding of graphics pipelines, but could be a good start.

Given that we don't use `state->stencil.reference` anymore, should this also go away?

Please feel free to take this as a starting point, or simply close it @RandyGaul.

### DIFF
--- a/src/cute_graphics.cpp
+++ b/src/cute_graphics.cpp
@@ -1678,6 +1678,25 @@ void cf_apply_scissor(int x, int y, int w, int h)
 	SDL_SetGPUScissor(s_canvas->pass, &scissor);
 }
 
+void cf_apply_stencil_reference(int reference)
+{
+  CF_ASSERT(s_canvas);
+  CF_ASSERT(s_canvas->pass);
+  SDL_SetGPUStencilReference(s_canvas->pass, reference);
+}
+
+void cf_apply_blend_constants(float r, float g, float b, float a)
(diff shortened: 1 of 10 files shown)


src/cute_graphics.cpp
```diff
@@ -1832,7 +1851,6 @@ static SDL_GPUGraphicsPipeline* s_build_pipeline(CF_ShaderInternal* shader, CF_R
 	pip_info.depthStencilState.frontStencilState.compareOp = s_wrap(state->stencil.front.compare);
 	pip_info.depthStencilState.compareMask = state->stencil.read_mask;
 	pip_info.depthStencilState.writeMask = state->stencil.write_mask;
-	pip_info.depthStencilState.reference = state->stencil.reference;
```
waldnercharles, line 1835: I think you need to call `SDL_SetGPUStencilReference` in `cf_apply_shader` to apply the stencil still.

pusewicz: Done in https://github.com/RandyGaul/cute_framework/pull/215/commits/b1cb661f67d38969ceb87beab824f744c313e5b7

PR #227, opened 2024-09-22
Do a non-shallow clone of SDL3

bullno1:
This fixes the current failed build in CI.

However, CI time significantly increases due to non-shallow clone.

The alternative is that you can fork SDL3 and only update it with upstream when needed.


pusewicz:
Another option is to point to a ZIP file eg. `https://github.com/libsdl-org/SDL/archive/88a01fbc964fbea1203071ea3ca2c6f799d4f571.zip`

bullno1:
I didn't know they provide archive for each commit. Wow.

PR #229, opened 2024-09-22
Make it possible to build as a dynamic library

bullno1:
Only tested on Linux so far but hot reload works.


RandyGaul:
Not sure about the fpic force for non-linux, I'll take a look at this at some point though. Thank you!

bullno1:
AFAIK, it is not a thing in Windows: https://stackoverflow.com/a/13309747.
So the setting should  just noop if it's clang.

In MSVC, there is no such flag anyway.

Mac is unix so it should be similar.

PR #236, opened 2024-09-23
Fix cf_text_size

bullno1:


bullno1:
Wait, this means y is negative.

src/cute_draw.cpp
```diff
@@ -2093,9 +2102,13 @@ static v2 s_draw_text(const char* text, CF_V2 position, int text_length, bool re
 		if (vertical) {
 			x += w;
 			y = initial_y;
+
+			max_x = fmaxf(max_x, x);
```
RandyGaul, line 2106: Use `min` and `max` for these

src/cute_draw.cpp
```diff
@@ -2093,9 +2102,13 @@ static v2 s_draw_text(const char* text, CF_V2 position, int text_length, bool re
...
 		} else {
 			x = position.x;
 			y -= line_height;
+
+			min_y = fminf(min_y, y + font->descent * scale);
```
RandyGaul, line 2111: Same

PR #237, opened 2024-09-23
Fix string manipulation functions

bullno1:
I was running my program through AddressSanitizer and these show up:

* cf_serase: `scount` already includes the null terminator so +1 is redundant.
* cf_sreplace: `remaining` is wrongly calculated based on `with_len` instead of `replace_len`. We want to overwrite the replaced substring with what follows after it so `replace_len` should be used in all cases.

The current code works in most cases since a new page is zero filled and accessing out of bound will just return 0.
But it will break if the string lies near the border of a page or the allocator recycle some memory chunks.


RandyGaul:
Awesome thanks! These little functions are really error prone so it's great to get more testing on them.

PR #240, opened 2024-09-26
Fix memory error when unloading easy sprite

bullno1:
* cf_easy_sprite_unload: This should free the image before removing the sprite entry. Otherwise, the image pointer would be invalidated.
* cf_get_pixels: Guard against invalid id when handling font and easy sprite. This is because after a resource is unloaded, its id still lingers in the batcher for a while.


RandyGaul:
Looks good, thank you! Nice catch.

PR #241, opened 2024-09-27
Improve IME and text input support.

bullno1:
* Allow grabbing the entire text input buffer.
  At least it is useful for my IME which sends an entire composed word.
* Replace rather than append composition.
  It is what the IME is "cooking".
  At least on Linux, the IME handles text replacement and the application just have to render.
* Add an IME example.
  Only tested with https://en.wikipedia.org/wiki/UniKey_(software) on Linux but SDL **should** be portable.

There are a few things that may need further thoughts:

Should `cf_draw_text` accepts `NULL` as input and treat it as an empty string?
CF allows initializing a string as `NULL` and manipulate it.
So `NULL` is equivalent to an empty string.

Do we need a way to render text from utf32?
The current example keeps 2 concurrent representations: utf8 and utf32.
utf32 is for editing since each codepoint is a glyph.
utf8 is for rendering since CF expects it but upon rendering, it is decoded into utf32 anyway.
That said, I'm not sure how text editors do this.
I only need short textboxes so it doesn't really matter that much.

What should `cf_text_size` return?
Previously, I changed it to be inclusive of the entire text box.
But to render a cursor on a multilline string, that is not enough.
For example:

```
First line is long
Second | <- cursor is here.
```

Maybe a `CF_TextInfo cf_text_info(const char* text, int num_chars)` where `CF_TextInfo` can accommodate different types of info.

Supporting IME composition candidates (think auto complete for IME): https://wiki.libsdl.org/SDL3/SDL_HINT_IME_IMPLEMENTED_UI#remarks.
It is apparently a thing but I have no knowledge of this nor do I need it soon.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -489,6 +489,7 @@ if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
 		add_executable(platformer samples/platformer.cpp)
 		add_executable(polygon samples/polygon.cpp)
 		add_executable(scissor samples/scissor.c)
+		add_executable(ime samples/ime.c)
 		set(SAMPLE_EXECUTABLES
 			easysprite
 			basicserialization
@@ -521,6 +522,7 @@ if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
 			platformer
 			polygon
 			scissor
+			ime
 		)
 
 		foreach(CURRENT_TARGET ${SAMPLE_EXECUTABLES})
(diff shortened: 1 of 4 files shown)


RandyGaul:
Looks very well written!

RandyGaul:
> Should cf_draw_text accepts NULL as input and treat it as an empty string?
Yes.

> Do we need a way to render text from utf32?
Probably not.

> What should cf_text_size return?
> Previously, I changed it to be inclusive of the entire text box.
Sounds correct to me.

> Maybe a CF_TextInfo cf_text_info(const char* text, int num_chars) where CF_TextInfo can accommodate different types of info.
This is a good direction. Let the user have more optional info about the text glyphs, and we can just pull it out from the font and hand it back. I don't want to clutter CF with anything too much or too specific. We want to offload more custom or nuanced stuff to the user.

PR #250, opened 2024-11-18
Fix logic for render command merging

bullno1:
I believe the logic is flipped.


RandyGaul:
I've a lot of issues with this chunk of code lately... Will take a closer look soon.

RandyGaul:
Going to just squash this and try it out directly.

PR #251, opened 2024-11-22
Make runtime shader compilation optional

bullno1:
Huge changes:

* Add libcute-shader as a wrapper for glslang.
* Add cute-shaderc, an offline compiler that can generate header for embedding.
* Add various functions to accept bytecode from generated headers.
* Precompile builtin shaders and also allow runtime compilation.
   Builtin shaders have to be declared in src/cute_shader/builtin_shaders.h
* Updated all samples to support both precompiled bytecode and runtime compilation.

As of this writing, windows build in CI is blocked by: https://github.com/actions/runner-images/issues/10978.

~~Allso, `#include` within the `cf_shader_directory` is not yet resolved.~~ (Implemented)
I'm posting this early for feedbacks.


CMakeLists.txt
```diff
@@ -466,18 +491,18 @@ if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
...
+		add_executable(shallow_water samples/shallow_water.cpp samples/shallow_water_data/shader.h)
 		add_executable(noise samples/noise.c)
 		add_executable(fetch_image samples/fetch_image.cpp)
-		add_executable(metaballs samples/metaballs.cpp)
+		add_executable(metaballs samples/metaballs.cpp samples/metaballs_data/shader_shd.h)
 		add_executable(timestep samples/timestep.cpp)
 		add_executable(joypad samples/joypad.c)
 		add_executable(outline_stencil samples/outline_stencil.cpp)
-		add_executable(recolor samples/recolor.cpp)
+		add_executable(recolor samples/recolor.cpp samples/recolor_data/recolor_shd.h)
```
RandyGaul, line 505: I would prefer the majority of sample code to use runtime string shaders. Precompiling is unneeded complexity when starting out.

include/cute_draw.h
```diff
@@ -1178,6 +1178,18 @@ CF_API CF_RenderState CF_CALL cf_draw_peek_render_state();
...
 CF_API void CF_CALL cf_draw_set_atlas_dimensions(int width_in_pixels, int height_in_pixels);
 
+/**
+ * @struct   CF_DrawShaderBytecode
+ * @category draw
+ * @brief    Bytecode for a draw shader.
+ * @remarks  This can be created using the `cute-shaderc` compiler.
+ * @related  CF_Shader cf_draw_push_shader cf_draw_pop_shader cf_draw_peek_shader cf_make_draw_shader_from_bytecode
+ */
+typedef struct CF_DrawShaderBytecode {
```
RandyGaul, line 1188: Newline on brace to match the rest of CF. We will need a page on CF website documenting `cute-shaderc` for sure. I can help get you going on that if you need -- it's really easy.

samples/hello_triangle_data/tri_fs.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+static const uint8_t s_tri_fs_bytecode_content[448] = {
```
RandyGaul, line 3: We want, in a comment, the string representation of the shader just above the bytecode here. Perhaps the compiler should output this directly into the header. sokol-shdc does this and it's greatly appreciated

src/cute_shader/cute_shader.cpp
```diff
@@ -0,0 +1,238 @@
...
+#define CUTE_SHADER_GLSL_VERSION 450
+
+namespace cute_shader {
+
+struct BuiltinInclude {
+	const char* content;
+	size_t size;
+};
+
+class Includer: public glslang::TShader::Includer {
```
RandyGaul, line 21: Interesting, is this includer thing a part of the glslang API? I hate it, but looks like it's just a way to use glslang, which is accpetable and I'm glad you hid this all back here in the implementation in an isolated place.

RandyGaul (commented):
Questions:
- Is the shader compiler a standalone program? Which source is it exactly? I didn't want to spend time trawling the cmake file to find out. Is it just cute_shaderc.cpp and associated header?
- Could you outline here what the major individual build targets are exactly and outline their dependencies to each other?
- Can the shader compiler be linked in statically into cute.lib or cute.dll (static or dynamic libs, but not as a standalone exe), as a simple cmake option?
- Is this easy for users to toggle on/off in cmake? I want runtime shaders in CF on by default, and behind a single cmake option to turn them off and NOT compile or pull in any SPIRV nonsense. Is that the way things work here as of now?
- Do you have plans yet to add in a dedicated website page for documenting the compiler?
- Can the compiler output serialized SPIRV blobs directly, and not just C headers?
- Can the compiler output reflection info required for passing into SDL_Gpu (various input counts for specific resource sets)?
- I'd prefer only one or two samples to use bytecode. The rest should just use runtime shaders alone to reduce complexity.

RandyGaul:
Would also be quite nice to document in the code as comment a rough plan on transitioning to SDL shader tools in the future, explaining how the current SPIRV handling is temporary until SDL evolves more thoroughly.

samples/hello_triangle_data/tri_fs.h
bullno1: Sure. Preprocessed shader or original code?

src/cute_shader/cute_shader.cpp
bullno1: Yes, it is glslang API to search for include file. It is passed to glslang:TShader::preprocess. It will call the includer to search for files.

I use that to implement lookup in the VFS, builtin shader include and also automatic include guard.

In the case of duplicated includes, the content is resolved into an empty string.

samples/hello_triangle_data/tri_fs.h
RandyGaul: Probably preprocessed

CMakeLists.txt
bullno1: Oh I made the changes just to check that all samples are fine with `CF_GLSLANG=OFF`

I can revert but those samples would crash when `CF_GLSLANG=OFF`, is that acceptable?

RandyGaul: Yep that's fine. As long as they all work with default settings when cloning, if the user turns it off I would expect them to then crash.

bullno1:
> Is the shader compiler a standalone program? Which source is it exactly? I didn't want to spend time trawling the cmake file to find out. Is it just cute_shaderc.cpp and associated header?

Yes, it is just `src/cute_shader/cute_shaderc.cpp`.

> Could you outline here what the major individual build targets are exactly and outline their dependencies to each other?

First, all glslang and spriv targets are now unconditionally fetched.

cute-shader is a library that wraps glslang compilation into spirv.
It depends on glslang.

cute-shaderc is a CLI program that can compile glsl to spriv and generate a C header.
It depends on cute-shader.
Notably it is also used to generate `src/data/builtin_shaders_bytecode.h` with the `-type=builtin` flag.
Mostly because I didn't want to make another program just for that.
This is done through a `add_custom_command` so whenever `src/cute_shader/builtin_shaders.h` is edited, `cute-shaderc` will get rebuilt and the bytecode header will be regenerated, keeping it in sync.

cute itself always depend on SPIRV tools since it still needs to do reflection even without runtime compilation.
The `cute-shader` dependency is conditional.

When `CF_GLSLANG=OFF` it will not link to `cute-shader`.
All compilation functions just return NULL.
Samples using inline string shader will crash (with a message in stderr).

When `CF_GLSLANG=ON` it will link to `cute-shader`.
All compilation functions will call `cute_shader_compile`.
Include now works with VFS and it will search in the `cf_shader_directory` if set.
The directory is passed to `cute-shader` as an include directory, similar to the `-I` flag in `cute-shaderc`.

> Can the shader compiler be linked in statically into cute.lib or cute.dll (static or dynamic libs, but not as a standalone exe), as a simple cmake option?

Yes. It is the `CF_GLSLANG` option.

> Is this easy for users to toggle on/off in cmake? I want runtime shaders in CF on by default, and behind a single cmake option to turn them off and NOT compile or pull in any SPIRV nonsense. Is that the way things work here as of now?

Not quite yet.
glslang is always pulled because the shader compiler requires that.
I can make it so that the shader compiler is not built.
Then when `CF_GLSLANG=OFF` and `CF_CUTE_SHADERC=OFF`, glslang will not be pulled.

SPIRV-tool is still needed regardless of options because I have only implemented bytecode compilation, not reflection.
I plan to do that next.
The current internal pipeline is: text -> bytecode -> reflection -> CF_Shader.
This only skips the `text -> bytecode` path.

> Do you have plans yet to add in a dedicated website page for documenting the compiler?

Yes, I can do that.

> Can the compiler output serialized SPIRV blobs directly, and not just C headers?

It was possible earlier but I reverted it because draw shader, the more common case is a couple of shaders.
This can be added back and it will error on cases other than `-type=vertex` or `-type=fragment`.
The other values are:

* `draw`: Draw shader, which is actually 2 fragment shaders.
* `builtin`: Dump all builtin shaders, for when runtime compilation is turned off.

> Can the compiler output reflection info required for passing into SDL_Gpu (various input counts for specific resource sets)?

Not yet.
That should be next in the list.

> I'd prefer only one or two samples to use bytecode. The rest should just use runtime shaders alone to reduce complexity.

Sure. I will only make changes to samples that already use shader in file and revert changes from samples with inline shaders.

bullno1:
> Would also be quite nice to document in the code as comment a rough plan on transitioning to SDL shader tools in the future, explaining how the current SPIRV handling is temporary until SDL evolves more thoroughly.

Tbh, I have no idea how SDL shader tool is heading.
This is my reference: https://github.com/libsdl-org/SDL_shader_tools.

I'll add some comment on how it is a temporary solution.
But I think what we can somewhat guarantee is that the API will remain the same.
That includes the compiler and all its flags, the include behaviour etc...

Shader syntax may change depending on what SDL_shader tool is doing.

RandyGaul:
> But I think what we can somewhat guarantee is that the API will remain the same.

Thanks! Excellent.

RandyGaul:
> cute itself always depend on SPIRV tools since it still needs to do reflection even without runtime compilation.

If we attach reflection info can we instead cut out SPIRV from CF runtime, unless online shader compilation is required? We can augment the bytecode blob with all necessary info. Thoughts?

bullno1:
> If we attach reflection info can we instead cut out SPIRV from CF runtime (other than SPIRV-Cross), unless online shader compilation is required? We can augment the bytecode blob struct with all necessary info. Thoughts? Just a few more structs to store extra reflection info for CF runtime, mainly uniform information.

Yes, that's the plan.
My previous comment was about the current state of the PR but I'm heading towards moving all Spirv dependencies (except for SPIRV-Cross) into `cute-shader`.

docs/topics/shader_compilation.md
```diff
@@ -0,0 +1,176 @@
...
+With this option enabled, the build process will involve pulling in [glslang](https://github.com/KhronosGroup/glslang).
+A Python 3.x installation is also required.
+CF's own builtin shaders are also compiled everytime the program starts.
+
+To remove the online compiler from CF, set `CF_RUNTIME_SHADER_COMPILATION` to `OFF`.
+Take note that online compilation functions such as [`cf_make_shader_from_source`](/graphics/cf_make_shader_from_source.md) will always fail if this is the case.
+
+## Offline compilation
+
+If you have a lot of shaders and want to avoid runtime compilation overhead, CF provides its own offline compiller: `cute-shaderc`.
```
RandyGaul, line 24: compiller typo

src/cute_shader/cute_shader.h
```diff
@@ -0,0 +1,78 @@
+#ifndef CUTE_SHADER_H
+#define CUTE_SHADER_H
+
+#include <stddef.h>
+#include <cute_shader_bytecode.h>
+
+typedef struct cute_shader_define_t {
```
RandyGaul, line 7: Could we use CF_ style for structs, and brace on separate line

include/cute_shader_bytecode.h
```diff
@@ -0,0 +1,203 @@
...
+} CF_ShaderUniformInfo;
+// @end
+
+/**
+ * @struct   CF_ShaderInputInfo
+ * @category graphics
+ * @brief    Information about an input of a vertex shader.
+ * @related  CF_ShaderBytecode
+ */
+typedef struct CF_ShaderInputInfo {
```
RandyGaul, line 126: Separate brace line

src/cute_shader/cute_shaderc.cpp
```diff
@@ -0,0 +1,589 @@
...
+#define HEADER_LINE_SIZE 16
+
+typedef enum {
+	SHADER_TYPE_VERTEX,
+	SHADER_TYPE_FRAGMENT,
+	SHADER_TYPE_DRAW,
+	SHADER_TYPE_BUILTIN,
+} shader_type_t;
+
+static const char* parse_flag(const char* arg, const char* flag_name) {
```
RandyGaul, line 27: Brace on different line

RandyGaul:
Looking quite good, and nice job on the documentation. Let me know when you feel like it's ready for me to pull it down and try things out.

src/cute_shader/cute_shader.h
bullno1: Sure, for cute shader, it's not really part of CF though. How should I name it:

* `CF_ShaderDefine`
* `CSH_Define`: CuteShader define.
* `CS_Define`: But `CS` is already `cute_sound`

bullno1:
Except for [stylistic changes](https://github.com/RandyGaul/cute_framework/pull/251/files/45e91155d0bc9016aa099a1c862be54f253d77a8#r1855365373), which I'll get to later today, there should not be anymore implementation changes.

src/cute_shader/cute_shader.h
RandyGaul: I'd say just go with CF_ prefix

RandyGaul:
You did a very solid job on this. Going to just merge it in and make any adjustments I think of in main branch.

RandyGaul:
If I want to update the CF shaders and regenerate the header, how were you going about doing this? Did you have a command line for that?

bullno1:
It's part of the build: https://github.com/RandyGaul/cute_framework/pull/251/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR433-R443

PR #254, opened 2024-11-28
Fix looping behaviour for pingpong animation

bullno1:
The sprite would loop properly in ping pong mode.

If loop is set to false, it will bounce back once and stop.


bullno1:
Wait, I get weird origin behaviour with this.

PR #255, opened 2024-11-28
Fix how sprite origin is loaded from aseprite

bullno1:
* Apply pivot from a slice's frame to the end of the sprite.
* Check that the slice is named "origin".

Reference: https://github.com/aseprite/aseprite/blob/main/docs/ase-file-specs.md#slice-chunk-0x2022

>   DWORD     Frame number (this slice is valid from this frame **to the end of the animation**)


bullno1:
> The slice named "origin"'s center is used to define the local offset.

Now that I read this again, I'm not sure what is correct.

On one hand, aseprite is arbitrary with the pivot feature and even slice.
It is meant to be "userdata".
So CF can define its own convention.

On the other, at least in the aseprite editor, the pivot is not relative to the center of the slice or has anything to do with the bounds at all.
It is always independent and the origin is always at the top left being `(0, 0)`.
And it always have that shenanigans where if the slice is the same for several frames, it decides to save only one frame as stated in the format spec.

RandyGaul:
This is a bit of a finnicky feature in aseprite, like we discussed, so I'll take a closer look at a later time once I get back into dev on my own game project here, and can give it the attention it deserves in an actual project.

PR #258, opened 2025-03-01
Set high density window property

pusewicz:
This fixes the HighDPI mode rendering on macOS.

https://wiki.libsdl.org/SDL3/SDL_CreateWindowWithProperties


RandyGaul:
Could you show before/after differences? I'm not sure exactly what the change is doing

pusewicz:
@RandyGaul I found a simpler way to fixing this problem. We were setting the flag on the `options` variable and then never doing anything with it. It should be set on the `flags` so that it does get passed to the window properties and set on the window like we do with the other flags.

### Before

Notice the scale being returned as `1`.

Screenshot 2025-03-02 at 08 32 55
Screenshot 2025-03-02 at 08 35 43

Cute was rendering everything at DPI of 1x and was upscaled to 2x by the OS making the edges of the pixels blurry.

### After

Notice the scale being returned as `2`, as expected.

Screenshot 2025-03-02 at 08 34 19
Screenshot 2025-03-02 at 08 35 48

RandyGaul:
Gotcha, thanks for the explanation!

PR #259, opened 2025-03-01
Framerate functions

pusewicz:
This adds 2 framerate functions.

@RandyGaul Let me know if you are ok with the naming.

### DIFF
--- a/include/cute_app.h
+++ b/include/cute_app.h
@@ -710,6 +710,22 @@ CF_API void CF_CALL cf_app_set_title(const char* title);
 CF_API void CF_CALL cf_app_set_icon(const char* virtual_path_to_png);
 
 /**
+ * @function cf_app_get_framerate
+ * @category app
+ * @brief    Returns the current framerate of the application.
+ * @related  cf_app_get_framerate cf_app_get_smoothed_framerate
+ */
+CF_API float CF_CALL cf_app_get_framerate();
+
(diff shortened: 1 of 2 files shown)


RandyGaul:
Looks good! Could we note in the docs it's over 10 frames for the smoothed variant?

pusewicz:
@RandyGaul Added!

BTW. Would this be an improvement if a local define was introduced to make it easier to change and refer to the same value?

```c
float cf_app_get_smoothed_framerate()
{
	#define FRAMES 10
	static float fps[FRAMES] = { 0 };
	static int frame_index = 0;
	fps[frame_index] = 1.0f / CF_DELTA_TIME;
	frame_index = (frame_index + 1) % FRAMES;
	float sum = 0;
	for (int i = 0; i < FRAMES; ++i) {
		sum += fps[i];
	}
	return sum / (float)FRAMES;
	#undef FRAMES
}
```

RandyGaul:
One way to introduce a variable is to store a float and lerp from previous to current based on the tunable. For example if want an average of last N frames we can do: `fps = lerp(prev, current, 1 / N)`. This way you don't need to store N elements, and N can be a tunable itself.

pusewicz:
@RandyGaul Cool, pushed a change to accommodate for that. I've also increased the smoothing to the last 60 frames.

PR #267, opened 2025-03-05
Include path in the error message

pusewicz:
When returning an error message from the `cf_image_load_png` and `cf_image_load_png_indexed`, let's include the path that was problematic, so it's faster to identify the problem.


bullno1:
The path is known by the caller right?

Can't the caller just check the error and log accordingly?

PR #268, opened 2025-03-07
Remove unused variable in the shaders

pusewicz:
This PR removes the unused variables set in the shaders.


RandyGaul:
I want to keep these for debugging. It's OK if they are left in as the compiler will simply strip them. If it's causing warnings that you may be seeing, commenting them out is better.

PR #269, opened 2025-03-07
Use SDL_WaitAndAcquireGPUSwapchainTexture to wait for an available texture

pusewicz:


RandyGaul:
I don't think this is a good change here, as there are some use-cases otherwise. It's more flexible to simply drop rendering when GPU bound and let the user decide how to proceed (e.g. waiting, or maybe something else). What CF could use is a predicate to say true/false for if we are CPU or GPU bound. For example `cf_app_is_cpu_bound` or `cf_app_is_gpu_bound`.

RandyGaul:
https://github.com/RandyGaul/cute_framework/issues/270

PR #272, opened 2025-03-10
Bump SDL to version 3.2.8

pusewicz:


pusewicz:
In general, given those are patch version updates, one would assume they have no breaking changes. And to me it feels like being up to date is better than doing a bigger jump later.

But sure, I'll follow your advice.

PR #273, opened 2025-03-12
Disable buffer cycling in SDL_UploadToGPUBuffer/SDL_UploadToGPUTexture

bullno1:
This leads to a significant reduction in VRAM usage.

According to https://moonside.games/posts/sdl-gpu-concepts-cycling/:

* `SDL_MapGPUTransferBuffer(cycle=true)` ensures that cycling happen if there is any in flight **upload** command. This is to ensure that uploaded data is not corrupted. This is why turning this off and without vsync creates visual corruption.
* `SDL_UploadToGPUBuffer/Texture(cycle=true)` however, ensures that cycling happens if the resource is bound to an earlier **render** command in the command buffer.
  This ensures that render data is not corrupted due to an in-flight upload.
* But because commands are ordered on the GPU due to barriers and the way CF uploads before rendering, I think cycling on upload is not needed.
  Cycling is only needed on mapping because it happens concurrently on both CPU and GPU.

I have personally tested between master (`37767848969574ba4d003d1504030c003e8edd67`) and this branch.
On master, the output of `nvidia-smi` for some samples are:

```
0   N/A  N/A           52465    C+G   ./textdrawing                          1389MiB
```

On this branch:

```
0   N/A  N/A           52998    C+G   ./textdrawing                           109MiB
```

There is no visual corruption in various samples after switching to this branch.

### DIFF
--- a/src/cute_graphics.cpp
+++ b/src/cute_graphics.cpp
@@ -229,7 +229,7 @@ void cf_texture_update(CF_Texture texture_handle, void* data, int size)
 	src.pixels_per_row = tex->w;
 	src.rows_per_layer = tex->h;
 	SDL_GPUTextureRegion dst = SDL_GPUTextureRegionDefaults(tex, tex->w, tex->h);
-	SDL_UploadToGPUTexture(pass, &src, &dst, true);
+	SDL_UploadToGPUTexture(pass, &src, &dst, false);
 	SDL_EndGPUCopyPass(pass);
 	if (!tex->buf) SDL_ReleaseGPUTransferBuffer(app->device, buf);
 	if (!app->cmd) SDL_SubmitGPUCommandBuffer(cmd);
@@ -853,7 +853,7 @@ static void s_update_buffer(CF_Buffer* buffer, int element_count, void* data, in
 	region.buffer = buffer->buffer;
 	region.offset = 0;
 	region.size = size;
-	SDL_UploadToGPUBuffer(pass, &location, &region, true);
+	SDL_UploadToGPUBuffer(pass, &location, &region, false);
 	SDL_EndGPUCopyPass(pass);
 	if (!app->cmd) SDL_SubmitGPUCommandBuffer(cmd);
 }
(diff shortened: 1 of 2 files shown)


bullno1:
But reading this again: https://moonside.games/posts/sdl-gpu-concepts-cycling/#data-dependencies

I think this does create inter-frame dependency.

bullno1:
Also, unlike the vertex buffer, the texture atlas is updated a lot less often so it can have cycling without dramatic increase usage in VRAM, at least in the existing samples.

I'm referring to this: https://github.com/RandyGaul/cute_framework/pull/273/files#diff-f51014e540a7ba51bd8c6511c352711a87a0e82aa8ff59b9eed7afc5f01eaebfR232

That said, most samples are fairly static.
But they are also small enough that frame dependency does not seem to matter.

RandyGaul:
Turning on cycling shouldn't be introducing high VRAM utilization if the FPS is limited to something low. If it is still, then there's some bugs. I was testing by limiting FPS to 10.

bullno1:
I tried `cf_set_target_framerate(10)` in the text drawing example and the VRAM usage is still high (1.3GB).
The frame rate is indeed low.

RandyGaul:
I won't have time to test/debug myself for a little while, do you recommend merging in the PR or is there remaining work to explore?

bullno1:
I think it's better to test on a second device/setup.

Could be nvidia on Linux being funny although I have a strong feeling it is the case.

pusewicz:
@bullno1 Do you want me to test on macOS/M3?

bullno1:
If you have time yeah.

bullno1:
@pusewicz can you try: https://github.com/bullno1/cute_framework/tree/pu-memory-issue-demo-fix

nvidia-smi reports only 109MiB for me.

bullno1:
This is much better: https://github.com/RandyGaul/cute_framework/pull/274

bullno1:
Reduction on the RAM side: https://github.com/RandyGaul/cute_framework/pull/274

The 2 problems seem to be orthogonal.

bullno1:
This seems to be a better fix: https://github.com/RandyGaul/cute_framework/pull/275

PR #275, opened 2025-03-14
Defrag the sprite batcher before every flush

bullno1:
This ensures that the atlas is defragged before every flush.
Both RAM and VRAM usage are reduced significantly.
No change to the cycling flag is needed.

However, I have no idea how this affects performance since there is another per-frame defrag.

### DIFF
--- a/src/cute_draw.cpp
+++ b/src/cute_draw.cpp
@@ -2819,6 +2819,7 @@ static void s_process_command(CF_Canvas canvas, CF_Command* cmd, CF_Command* nex
 		// Process the collated drawable items. Might get split up into multiple draw calls depending on
 		// the atlas compiler.
 		draw->need_flush = false;
+		spritebatch_defrag(&draw->sb);
 		spritebatch_flush(&draw->sb);
 	}
 }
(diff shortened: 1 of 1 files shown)


bullno1:
This seems to work but I think it should be a dirty flag instead?

Like:

* If there is a new sprite, mark the batcher as dirty.
* If there is a flush, defrag if dirty and clear the dirty flag.
* Per frame, defrag on tick if not already.

pusewicz:
This branch brings my memory usage in the demo to normal values (650MB) which I think is fine (it always seems to be more on macOS). The CPU is on ~100% though.

bullno1:
If you turn on vsync (`cf_app_set_vsync(true)`), does CPU usage drop?

Also, I suspect that calling defrag so often is suboptimal.

pusewicz:
Much better, but still it's higher at around ~30%.

Here's htop:

image

Here's my Activity Monitor:

image

But you know what, even on my branch the CPU usage is the same. Also even with vsync false I see it at ~65%. I guess depends what display I run it on.

To sum up, the CPU usage seems to be the same before and after the change.

pusewicz:
I don't think the thread count is an issue. But again, I think this change works.

RandyGaul:
Defrag is designed to be called per frame

bullno1:
So calling it before every flush is fine?

RandyGaul:
Yes. There was an issue with imgui though, needing to defrag/flush after drawing, but there's an if-statement and dirt-flag for this already.

RandyGaul:
Did you test with imgui?

bullno1:
Only the samples but none of them render images

RandyGaul:
I think I'm going to just merge this in. It's fine to continue with tweaks or fixes moving forward if-needed, but this overall seems like an improvement, so no point in leaving it as a PR, imo.

PR #277, opened 2025-03-24
Add a demo game image to README

pusewicz:
I think it would be neat to have more demoable stuff presented in the README to show what the framework is about. This could bring in more people. Let me know what you think @RandyGaul!


RandyGaul:
Agreed!

PR #279, opened 2025-05-07
Mark enums as closed

pusewicz:
Marking these as either open or closed allows to import these enums as actual enums in Swift, rather than structs. This will make for better type checking and does not impact actual C/C++ code.

https://github.com/swiftlang/swift-evolution/blob/main/proposals/0192-non-exhaustive-enums.md

### DIFF
--- a/include/cute_app.h
+++ b/include/cute_app.h
@@ -49,7 +49,7 @@ typedef enum CF_DisplayOrientation
 	#define CF_ENUM(K, V) CF_##K = V,
 	CF_DISPLAY_ORIENTATION_DEFS
 	#undef CF_ENUM
-} CF_DisplayOrientation;
+} CF_CLOSED_ENUM CF_DisplayOrientation;
 
 typedef uint32_t CF_DisplayID;
 
@@ -211,7 +211,7 @@ typedef enum CF_AppOptionFlagBits
 	#define CF_ENUM(K, V) CF_##K = V,
 	CF_APP_OPTION_DEFS
 	#undef CF_ENUM
-} CF_AppOptionFlagBits;
+} CF_CLOSED_ENUM CF_AppOptionFlagBits;
 
 /**
  * @function cf_make_app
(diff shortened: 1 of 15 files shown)


RandyGaul:
Can you help educate me on what this whole open vs closed enum thing is? This is a clang specific feature?

For binding Lua I actually just used the macros from the headers to import and define my own data as-needed, e.g.

```cpp
#define CF_ENUM(K, V) WHATEVER_YOU_NEED
CF_DISPLAY_ORIENTATION_DEFS
#undef CF_ENUM
```

This way you can populate enums, or arrays, or convert to strings, etc. in any way you would like.

image

I think it may be preferable to keep macro-bloat out of CF headers in order to keep CF simple over the years.

pusewicz:
> This is a clang specific feature?

No, it's an attribute feature that's supported by both Clang and GCC.

In Swift, when importing C code, this gives a hint to the Swift compiler on how to import this enum. Because it's marked as a `enum_extensibility` (does not matter whether open or closed), this C enum gets imported as a native Swift enum, too. Without that annotation, it gets imported as a struct. This is less than ideal for type checking as some shenanigans have to happen (eg. I need to manually re-declare values, rather than using enum values).

As for open or closed, from what I understand, is to define whether the enums can change in future (for ABI compatibility, or something, this detail escapes me).

Some extra context can be found here: https://belkadan.com/blog/2021/10/Swift-Regret-Unannotated-C-Enums/

pusewicz:
@RandyGaul With your Lua bindings, that's doable as you are using... C++. Swift can only import headers as-is. Unless you are saying that I should define a _new_ enum in a custom header that would be used instead of the original?

RandyGaul:
Right so it would potentially be a new enum. It's just that these markups aren't standard C/C++ and in the long it's important to try and keep the CF source (especially the platform specific macros) as simple and minimal as possible.

PR #281, opened 2025-05-23
Upgrade to IMGUI 1.91.9b-docking

pusewicz:
This PR upgrades IMGUI to the newest `v1.91.9b-docking`. I ripped out the custom GPU renderer and used the supplied `sdlgpu3`. This simplifies the code greatly, as can be seen by the tiny `cute_imgui.cpp` file.

The new implementation fixes the issue I was having on macOS with a high-DPI display, where the UI would be rendered at 1x and in 1/4 of the screen.

> [!IMPORTANT]
> I have no experience working with GPU related stuff, this is all taped together by what I could figure out looking at the example code from Imgui and current implementation in the Cute integration code. This definitely needs some extra 👀 to verify it's updated correctly and does not do anything stupid.

#### Quick demo

https://github.com/user-attachments/assets/253b5d2b-cee1-4f79-81dc-1cd789e3030b

If this gets merged, I'll work on pulling Imgui/Cimgui as CMake deps so that it's easier to upgrade in future.


RandyGaul:
Is this ready for merging?

pusewicz:
Yep.

PR #288, opened 2025-06-05
Upgrade shadercross

pusewicz:
This is a WIP that upgrades the shadercross library.

TODO:

- [ ] Shadercross source should be added conditionally


RandyGaul:
You can just incorporate shadercross unconditionally. It loads dynamic libraries if-needed, so there's no other hard dependency by simply compiling in shadercross. I also wouldn't want to have any more conditional complexity on the cmake side.

pusewicz:
Fair enough.

pusewicz:
@RandyGaul Resolved the conflict and marked the PR as ready.

PR #299, opened 2025-07-24
Mkdocs for building docs

pusewicz:
This PR changes the doc build tool to [MkDocs](https://www.mkdocs.org) with the [Material theme](https://squidfunk.github.io/mkdocs-material/).

image

image

#### TODO

- [x] Remove checked in files that will get generated automatically https://github.com/RandyGaul/cute_framework/pull/300
- [x] Getting CMake build just the docsparser to generate the API Reference md files
- [x] Going through the code to remove the incorrect links
- [x] Setting up GitHub action to build the docs on merge to master


pusewicz:
@RandyGaul This is now ready for review, I would suggest merging https://github.com/RandyGaul/cute_framework/pull/300 first, so that all the big file removal changes can go away from this PR and it's easier to see the actual changes. I did test the build process, what's missing is testing of the upload to GitHub pages, which can only happen when we are happy with the changes and this PR gets merged.

When this PR lands, I'll spend some time tweaking the docs, from the content perspective, and updating to the style MkDocs supports (eg. different notes etc) and tweak plugins, so that eg. dark mode gets enabled.

PR #311, opened 2025-08-01
Fix imgui image

ogam:
Imgui SDL_GPU backend wants `SDL_GPUTextureSamplerBinding` instead of `SDL_GPUTexture` 
Spritebatcher was defragging during `cf_render_*_to()` so texture was being thrashed before imgui was able to send anything to SDL_GPU


RandyGaul:
Debug bit accidentally left on in the sample. Nice job tracking down the bug!

PR #315, opened 2025-08-19
font debug sample

ogam:
image


ogam:
those lines are probably still hard to see so would probably need to adjust it a bit

PR #317, opened 2025-08-19
Use `(void)` for nullary functions

bullno1:
Because in C: `f()` actually means `f(...)` and even though the calling convention makes it OK, it's weird to have no errors for: `cf_draw_pop_color(color)`.


bullno1:
> Q: Isn't there any kind of linter/compiler flag that would catch this?

Yes if you enable warning and warning as error like I do.

The thing is: cute_framework is in C++ so it's valid.
It's only a problem in C.

PR #318, opened 2025-08-21
Fix cf_app_set_icon

bullno1:
Confirmed on both Windows and Linux that the icon will show up with the correct color after this.


bullno1:
A separate problem whether this struct is wrong: https://github.com/RandyGaul/cute_framework/blob/master/include/cute_color.h#L50

PR #328, opened 2025-08-26
Pin spirv-cross

pusewicz:
Avoid automatic updates with UPDATE_DISCONNECTED


bullno1:
Does it have a tag? github (and I think git in general) does not let you pull from a hash that is not the tip of a branch or tagged?

pusewicz:
@bullno1 You are right, but I think only when [shallow clone](https://cmake.org/cmake/help/latest/module/ExternalProject.html#git) is used. I'll remove it, so it should just work.

PR #332, opened 2025-08-27
Store binding slots

### LINKED ISSUE #324: Texture binding slots issue
Currently if you write a shader and then comment out one of the textures, the slot bindings will be mismatched/counted by CF's implementation. For example:

```glsl
layout(set = 2, binding = 1) uniform sampler2D tex1;
layout(set = 2, binding = 2) uniform sampler2D tex2;
layout(set = 2, binding = 3) uniform sampler2D tex3;
```

And then comment one out:

```glsl
layout(set = 2, binding = 1) uniform sampler2D tex1;
//layout(set = 2, binding = 2) uniform sampler2D tex2;
layout(set = 2, binding = 3) uniform sampler2D tex3;
```

The third texture will get "missed" by CF's implementation. A proper solution would probably require a refactor of `CF_ShaderInternal` and potentially the glslang shader reflection to pull out shader slot bindings and store them in our own reflection structs.

This chunk of code can then use the new slot index instead of `j` index for assignments:

```cpp
	// Bind images to all their respective slots.
	int sampler_count = shader->image_names.count();
	SDL_GPUTextureSamplerBinding* sampler_bindings = SDL_stack_alloc(SDL_GPUTextureSamplerBinding, sampler_count);
	int found_image_count = 0;
	for (int i = 0; found_image_count < sampler_count && i < material->fs.textures.count(); ++i) {
		const char* image_name = material->fs.textures[i].name;
		for (int j = 0; j < shader->image_names.size(); ++j) {
			if (shader->image_names[j] == image_name) {
				sampler_bindings[j].sampler = ((CF_TextureInternal*)material->fs.textures[i].handle.id)->sampler;
				sampler_bindings[j].texture = ((CF_TextureInternal*)material->fs.textures[i].handle.id)->tex;
				found_image_count++;
			}
		}
	}
	CF_ASSERT(found_image_count == sampler_count);
	SDL_BindGPUFragmentSamplers(pass, 0, sampler_bindings, (Uint32)found_image_count);
```

pusewicz:
Closes #324

### DIFF
--- a/src/cute_shader/cute_shader.cpp
+++ b/src/cute_shader/cute_shader.cpp
@@ -526,7 +526,7 @@ void cute_shader_free_result(CF_ShaderCompilerResult result)
 		free((char*)shader_info->image_names[i]);
 	}
 	free(shader_info->image_names);
-	free(shader_info->image_binding_slots);
+	free((int*)shader_info->image_binding_slots);
 
 	free((void*)result.bytecode.content);
 	free((char*)result.preprocessed_source);
--- a/include/cute_shader_bytecode.h
+++ b/include/cute_shader_bytecode.h
@@ -153,7 +153,7 @@ typedef struct CF_ShaderInfo
 	/* @member Name of each images. */
 	const char** image_names;
 	/* @member Binding slot index of each image. */
-	int* image_binding_slots;
+	const int* image_binding_slots;
 
 	/* @member Number of uniform blocks. */
 	int num_uniforms;
--- a/src/data/builtin_shaders_bytecode.h
+++ b/src/data/builtin_shaders_bytecode.h
@@ -265,7 +265,7 @@ static const uint8_t s_draw_vs_bytecode_content[3176] = {
     0xFD, 0x00, 0x01, 0x00, 0x38, 0x00, 0x01, 0x00,
 };
 static const char** const s_draw_vs_bytecode_image_names = NULL;
-static int* const s_draw_vs_bytecode_image_binding_slots = NULL;
+static const int* const s_draw_vs_bytecode_image_binding_slots = NULL;
 static CF_ShaderUniformInfo* const s_draw_vs_bytecode_uniforms = NULL;
 static CF_ShaderUniformMemberInfo* const s_draw_vs_bytecode_uniform_members = NULL;
 static CF_ShaderInputInfo s_draw_vs_bytecode_inputs[14] = {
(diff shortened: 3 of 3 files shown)


src/cute_shader/cute_shader.cpp
```diff
@@ -432,6 +435,8 @@ CF_ShaderCompilerResult cute_shader_compile(const char* source, CF_ShaderCompile
 		if (num_images > 0) {
 			image_names = (const char**)malloc(sizeof(char*) * num_images);
 			memcpy(image_names, image_names_vec.data(), sizeof(char*) * num_images);
+			image_binding_slots = (int*)malloc(sizeof(int) * num_images);
```
bullno1, line 438: This should be freed in `cute_shader_free_result`

bullno1, line 438: This should be freed in `cute_shader_free_result`

src/cute_shader/cute_shaderc.cpp
```diff
@@ -106,8 +106,14 @@ static bool write_bytecode_struct(
 			fprintf(file, " \"%s\",", shader_info->image_names[i]);
 		}
 		fprintf(file, "\n};\n");
+		fprintf(file, "static int %s%s_image_binding_slots[%d] = {\n   ", var_name, suffix, shader_info->num_images);
```
bullno1, line 109: `const int` for consistency with the rest.

Also some **C** compilers (e.g: Clang) are strict about what is considered "constant expression" for static initializer used in `CF_ShaderBytecode` later

src/cute_shader/cute_shaderc.cpp
```diff
@@ -106,8 +106,14 @@ static bool write_bytecode_struct(
...
+		for (int i = 0; i < shader_info->num_images; ++i) {
+			fprintf(file, " %d,", shader_info->image_binding_slots[i]);
+		}
+		fprintf(file, "\n};\n");
 	} else {
 		fprintf(file, "static const char** const %s%s_image_names = NULL;\n", var_name, suffix);
+		fprintf(file, "static int* const %s%s_image_binding_slots = NULL;\n", var_name, suffix);
```
bullno1, line 116: `const inst* const` for the same reason.

src/cute_shader/cute_shader.cpp
```diff
@@ -520,6 +526,7 @@ void cute_shader_free_result(CF_ShaderCompilerResult result)
 		free((char*)shader_info->image_names[i]);
 	}
 	free(shader_info->image_names);
+	free((int*)shader_info->image_binding_slots);
```
bullno1, line 529: So the cast to free is not necessary if it wasn't const in the struct in the first place.

PR #338, opened 2025-09-07
Correctly declare function with no arguments

pusewicz:


RandyGaul:
lol that isn't "correctly", it's a difference from C and C++

pusewicz:
Understood. Thanks for the info. Was not aware of that.

PR #343, opened 2025-09-14
Use up to 4 threads

pusewicz:
This should help with OOMs


bullno1:
Github is stingy and C++ compilation consumes a lot of memory: https://github.com/RandyGaul/cute_framework/actions/runs/17716556220/job/50342287025?pr=343#step:5:1051 😂

PR #353, opened 2025-09-24
Ensure that text effects are cleared at the end of text rendering

bullno1:
Without this, the entire text will be affected when `text_length` is less than the actual sanitized length.

I'm not sure what should be done with the effect and the markup callbacks because rendering did not reach the end.


RandyGaul:
Doesn't it make sense to simply not call more callbacks for invisible text?

RandyGaul:
This one https://github.com/RandyGaul/cute_framework/blob/997c8b86d28261f06518d17e4631c9e741122738/include/cute_draw.h#L871

PR #359, opened 2025-09-29
Make basic_sprite sample compile and (partially) run on web

bullno1:
* Always link and init SPIRV-Cross
* Define away some pixel formats
* Use `emscripten_set_main_loop` for now


bullno1:
Wait, I think the texture define might be wrong

Issue #364, opened 2025-10-03
Update docs on ImGui to use ImGui_ instead of iG prefix


RandyGaul:
We swapped imgui bindings and all the docs about imgui still reference the old prefix using `iG`, so this needs to be updated to `ImGui_`.

PR #365, opened 2025-10-03
ImGui docs update

pusewicz:
I went ahead and updated the documentation. I also simplified some of the paragraphs to cut on words and increase signal-to-noise a little bit.

Closes https://github.com/RandyGaul/cute_framework/issues/364.


RandyGaul:
This looks really good!

PR #401, opened 2025-10-14
Draw to texture fix

pusewicz:
This is the attempt to fix the issues with the samples on macOS.

I saw that in the `s_draw_report`, after calling `cf_draw_elements()`, we also call `cf_commit()`. Following that approach in the `s_blit()` makes the `draw_to_texture` work:

image

Opening this as draft, as I'm not sure this is the correct way to fix this problem, and I'm not sure what the implications are.

Additionally, the other shatter demos have the same issue. I wonder if `cf_commit()` should be part of `cf_draw_elements()`?

### DIFF
--- a/docs/topics/low_level_graphics.md
+++ b/docs/topics/low_level_graphics.md
@@ -38,7 +38,6 @@ for each canvas {
             }
         }
     }
-    cf_commit();
 }
 ```
 
@@ -48,7 +47,6 @@ The important functions are are the apply functions. Each apply function is used
 - [`cf_apply_mesh`](../graphics/cf_apply_mesh.md)
 - [`cf_apply_shader`](../graphics/cf_apply_shader.md)
 - [`cf_draw_elements`](../graphics/cf_draw_elements.md)
-- [`cf_commit`](../graphics/cf_commit.md)
 
 ## Meshes
 
--- a/docs/topics/renderer.md
+++ b/docs/topics/renderer.md
@@ -359,7 +359,6 @@ CF renders through `SDL_Gpu`(https://wiki.libsdl.org/SDL3/CategoryGPU), a well-w
  *                 for each shader {
  *                     cf_apply_shader(shader, material);
  *                     cf_draw_elements(...);
- *                     cf_commit();
  *                 }
  *             }
  *         }


RandyGaul:
Seems like we should perhaps merge draw elements and commit together. Historically commit came from copying sokol_gfx's API design and just mirroring it 1:1 without much independent design consideration.

I rather like @bullno1's suggestion:

> `cf_draw_elements` also feels like it's basically forced since it's the only call that makes sense after `cf_apply_shader`. What if instead of `cf_apply_mesh(mesh)`, it's `cf_draw_elements(CF_Mesh* mesh)` after `cf_apply_shader`?

Putting a little more thought into it, commit at this time is acting like "I'm done with this pass" which is mostly what it did in sokol_gfx (sometimes it would just no-op depending on backend).

```c
void cf_sdlgpu_commit()
{
	SDL_EndGPURenderPass(g_ctx.canvas->pass);
}
```

I do agree the weirdness seems to be draw elements is separated from mesh. Though, in the case of static meshes it is entirely possible to want to apply a mesh once and then swap textures or maybe even shaders for different instances, but, perhaps that's just not worth an entire other API call.

`commit` typically means end of frame, so I don't think we should use that name. Perhaps a better name is just `cf_submit` or something like that. It just means the end of the pass, or in GLSL land it means clear all the dumb state I just set to render my prior stuff.

In CF the "end of frame" that a `commit` style API would usually cover is automagically handled in `cf_app_draw_onto_screen` even when using the low level cute_graphics.h header directly, basically handled by our own blit function and end frame function, e.g. `cf_gles_end_frame` or `cf_gles_blit_canvas`. So we don't need a public `commit`.

---

My tentative rec (would appreciate comments from others):
- Nix `cf_commit`
- Nix `cf_apply_mesh`
- Send in `CF_Mesh*` to `cf_draw_elements`
- Move old `cf_commit` code to end of `cf_draw_elements`

Thoughts?

pusewicz:
@RandyGaul The nixing of the `cf_apply_mesh` is more invasive, so going to hold on that. We can continue the discussion in https://github.com/RandyGaul/cute_framework/pull/403.

PR #403, opened 2025-10-15
WIP: Drop cf apply mesh

pusewicz:
This needs a refactor that introduces `cf_render` rather than `cf_apply_shader`.

### DIFF
--- a/docs/topics/renderer.md
+++ b/docs/topics/renderer.md
@@ -307,7 +307,7 @@ And that's for *all shape types*:
 
 # Text Rendering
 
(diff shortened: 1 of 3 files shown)


pusewicz:
I know it’s a noob question, but what are scissors for? Is that for cropping some textures to only display parts of them?Sent from my iPhoneOn 15 Oct 2025, at 10:32, Bach Le ***@***.***> wrote:﻿bullno1 left a comment (RandyGaul/cute_framework#403)
Another question is: What is the scope of states like scissor and viewport? Even in SDL_GPU, SDL_SetGPUScissor seems to imply that it affects the next draws onward until the pass is over but I haven't tested.
I don't really use viewport but scissor is how GUI libraries implement scrolling containers.

RandyGaul:
I think this PR is no longer needed

PR #407, opened 2025-10-17
Fix missing NUL terminator when copying GLSL source

### LINKED ISSUE #406: GLSL compile error when using an OpenGL backend
Hi. I'm using Cute framework with an OpenGL backend. On application starts up, the console prints multiple GLSL compilation error message.
```
Loaded GLES 3.2
GLSL compile error:
0(58) : error C0000: syntax error, unexpected $undefined at token "<undefined>"

GLSL compile error:
0(411) : error C0000: syntax error, unexpected $undefined at token "<undefined>"

GLSL link error:
Vertex info
-----------
0(58) : error C0000: syntax error, unexpected $undefined at token "<undefined>"
(0) : error C2003: incompatible options for link

Fragment info
-------------
0(411) : error C0000: syntax error, unexpected $undefined at token "<undefined>"
(0) : error C2003: incompatible options for link

GLSL compile error:
0(23) : error C0000: syntax error, unexpected $undefined at token "<undefined>"

GLSL compile error:
0(60) : error C0000: syntax error, unexpected $undefined at token "<undefined>"

GLSL link error:
Vertex info
-----------
0(23) : error C0000: syntax error, unexpected $undefined at token "<undefined>"
(0) : error C2003: incompatible options for link

Fragment info
-------------
0(60) : error C0000: syntax error, unexpected $undefined at token "<undefined>"
(0) : error C2003: incompatible options for link
```

Here is the code to repro the issue:
```C
#include <cute.h>

int main(int argc, char* argv[])
{
	CF_AppOptionFlags opt = CF_APP_OPTIONS_WINDOW_POS_CENTERED_BIT | CF_APP_OPTIONS_GFX_OPENGL_BIT;
	CF_Result result = cf_make_app("title", 0, 0, 0, 640, 480, opt, argv[0]);
	if (cf_is_error(result)) return -1;

	while (cf_app_is_running()) {
		cf_app_update(NULL);
		cf_draw_circle2(cf_v2(0.f, 0.f), 32.f, 2.f);
		cf_app_draw_onto_screen(true);
	}

	cf_destroy_app();
	return 0;
}
```

The compilation error occurs in [s_compile_shader](https://github.com/RandyGaul/cute_framework/blob/0dbcb04bc247c20e2f5c62de13880b8e90b84d54/src/cute_graphics_gles.cpp#L602-L615). When we are calling [glShaderSource](https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glShaderSource.xhtml), we passed NULL as the length parameter. This means the shader source string is assumed to be null terminated.
But `src` isn't null terminated, so OpenGL may just read past the buffer's boundry and encounter some invalid character, causing the compilation error.
```C++
static GLuint s_compile_shader(GLenum stage, const char* src)
{
	GLuint s = glCreateShader(stage);
	glShaderSource(s, 1, &src, NULL);
	glCompileShader(s);
	GLint ok = GL_FALSE;
	glGetShaderiv(s, GL_COMPILE_STATUS, &ok);
	if (!ok) {
		char log[4096]; GLsizei len=0;
		glGetShaderInfoLog(s, sizeof(log), &len, log);
		fprintf(stderr, "GLSL compile error:\n%.*s\n", (int)len, log);
	}
	return s;
}
```

In `cute_shader_compile` where the GLSL shader source is created, [the code](https://github.com/RandyGaul/cute_framework/blob/0dbcb04bc247c20e2f5c62de13880b8e90b84d54/src/cute_shader/cute_shader.cpp#L449-L452) allocates `glsl300_src` with size `strlen(source)`, and copies bytes from `source` to `glsl300_src`. But `strlen` doesn't include the null terminator. So `glsl300_src` is missing the null terminator '\0' and is later passed to `glShaderSource`.
```C++
		glsl300_src_size = strlen(source);
		glsl300_src = (char*)malloc(glsl300_src_size);
		memcpy(glsl300_src, source, glsl300_src_size);
		spvc_context_release_allocations(spvc);
```

t0byn:
Fixes #406

### DIFF
--- a/src/cute_shader/cute_shader.cpp
+++ b/src/cute_shader/cute_shader.cpp
@@ -447,8 +447,9 @@ CF_ShaderCompilerResult cute_shader_compile(const char* source, CF_ShaderCompile
 		}
 
 		glsl300_src_size = strlen(source);
-		glsl300_src = (char*)malloc(glsl300_src_size);
+		glsl300_src = (char*)malloc(glsl300_src_size + 1);
 		memcpy(glsl300_src, source, glsl300_src_size);
+		glsl300_src[glsl300_src_size] = '\0';
 		spvc_context_release_allocations(spvc);
 	}
 	spvc_context_destroy(spvc);


RandyGaul:
Ah!! This explains so much. Ty!!

bullno1:
That was actually intentional because there is an explicit length in the generated data.
The problem is the backend didn't use the length.

PR #408, opened 2025-10-18
Trigger trap on assertion failure

pusewicz:
This ensures that we use `__builtin_trap()` to trigger the debug break on other compilers like GNU cc and Clang.

https://clang.llvm.org/docs/LanguageExtensions.html#builtin-trap
https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005ftrap


RandyGaul:
Yes go ahead and add in the fancy debug break. It looks nice.

PR #437, opened 2026-01-02
Bump SDL to 3.4.0

pusewicz:
Examples work as expected.


bullno1:
What happened that blows up the dependency list?

pusewicz:
I took the list of deps from libsdl, as it was complaining about the lack of xscrnsaver. So rather than going back and forth and trying to figure out which dep to add, I used the same list SDL provides.

PR #444, opened 2026-01-18
Add IMGUI texture sample

theopechli:
image


RandyGaul:
Odd error: undefined reference to `imgui_texture_ref_from_id'

PR #458, opened 2026-02-05
Use CCache if available

pusewicz:


RandyGaul:
What is ccache and the idea here? Is this just for CI? Or is it intended to help local builds somehow?

pusewicz:
Ccache is mostly useful for CI (we already use it in CI for Linux/macOS). I want to make Windows builds use it too. If we enable directly in our CMakeLists.txt, it's going to be easier, and also will help with local builds, a bit. Mostly for games that are big, though.

PR #461, opened 2026-02-06
Use Ninja on Windows

pusewicz:
This makes it consistent across platforms and still uses the correct compiler. Might have some speed benefits, too.

`CMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded` is necessary so that when compiling with MSVC we actually do get hits in cache. Otherwise there's 0% hit-rate.


pusewicz:
@RandyGaul This is now ready.

PR #466, opened 2026-02-07
Docsparser

pusewicz:
Now that docsparser is not dependent on cute itself and its dependencies, we can extract a separate docs job, that only builds the docsparser binary and generates the docs when merging into master.


RandyGaul:
Conflicts

PR #474, opened 2026-02-13
Web samples

pusewicz:


RandyGaul (changes requested):
cmake step to generate shaders rather than checking them into git

bullno1:
> cmake step to generate shaders rather than checking them into git

The problem is: Web build has cute-shaderc disabled.
And even when enabled, it's a wasm binary.

Without it, we can't generate headers.

It is possible to:

* Build cute-shaderc from a different branch and pull it into this build with `depend:` or something.
* Generate header in desktop build and pull them here

RandyGaul:
By the way I don't think the sample building for web needs to be on every PR? I'm just a little worried that we're building way too much for CI? I don't want to run out of free service. Perhaps it can just be a manual trigger that we can run when a PR pops up to add samples (infrequent)

PR #476, opened 2026-02-13
Add emscripten build

pusewicz:
Apart from normal desktop builds, let's also test Emscripten.


pusewicz:
@RandyGaul This now builds!

PR #477, opened 2026-02-17
Introduce cf_sprite_is_valid

pusewicz:
`cf_sprite_is_valid` returns a `bool` depending whether the checks confirm the sprite is a valid struct by comparing the ids.

### DIFF
--- a/include/cute_sprite.h
+++ b/include/cute_sprite.h
@@ -24,6 +24,8 @@
 extern "C" {
 #endif // __cplusplus
 
+#define CF_SPRITE_ID_INVALID (~0ULL)
+
 /**
  * @enum     CF_PlayDirection
  * @category sprite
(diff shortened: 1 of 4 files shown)


RandyGaul:
I don’t really see how it’s safer and we can document a macro the same way we document anything else using docsparser. Macro would also be more performant.

Hmm macro can also work with all CF types, e.g. `#define CF_IS_VALID(handle) (handle.id)`, and _Generic can be used to specialize the sprite path. This would work with any of the very numerous handle types throughout CF and reduce the number of different docs pages we’d have to generate if we made a function for every single different kind of handle in all of CF. It would also not be as good of an experience to have to try and remember which is_valid function you need for each different handle.

RandyGaul:
Example: https://github.com/RandyGaul/cute_framework/commit/05df187a6b47c8c454e9b35563ed6605bee5b2f4

PR #495, opened 2026-03-11
Add TextEffect compositing

### LINKED ISSUE #480: text effect composition bug
When stacking multiple text effects only the first closed inner one affects the the text

Here I'm expecting a red color text, with a strike through and shaking
```c
cf_draw_text("<color color=#FF0000FF><shake><strike>Some text</strike></shake></color>", cf_v2(0, 0), -1);
```
but I'm seeing the following
Image

Adjusting the effect stack so color is the first closed one
```c
cf_draw_text("<color color=#FF0000FF><shake><strike>Some text</color></strike></shake>", cf_v2(0, 0), -1);
```
Image

waldnercharles:
**Text Effect Rendering**
* The strike-through effect (`<strike>`) now uses a per-glyph approach, storing the thickness and deferring the actual line drawing until after all effects and color changes are applied. This ensures the strike line matches the final rendered color and position, including gradients and alpha.
* The `CF_Strike` struct now includes a `color` field, allowing strike-through lines to use the correct color, including gradients.
* The code for spawning text effects can now handle multiple effects at the same character index.

Fixes #480

### DIFF
--- a/src/cute_draw.cpp
+++ b/src/cute_draw.cpp
@@ -2829,8 +2829,8 @@ static v2 s_draw_text(const char* text, CF_V2 position, int text_length, bool re
 				if (effect->strike_thickness > 0) {
 					float y_mid = y + h * 0.25f + (q0.y - pre_fx_q0_y);
 					CF_Strike strike;
-					strike.p0 = V2(q0.x, y_mid);
-					strike.p1 = V2(q0.x + xadvance, y_mid);
+					strike.p0 = V2(x, y_mid);
+					strike.p1 = V2(x + xadvance, y_mid);
 					strike.thickness = effect->strike_thickness;
 					strike.color = use_corner_colors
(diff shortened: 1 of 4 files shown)


src/cute_draw.cpp
```diff
@@ -2567,8 +2562,10 @@ static v2 s_draw_text(const char* text, CF_V2 position, int text_length, bool re
 	}
 
 	if (render || markups) {
+		if (!effect_state->alive) {
```
bullno1, line 2565: I don't understand.

Is this included so that the elapsed time is only ticked once per frame?

waldnercharles: Yeah, that's exactly right. Without it, an effect's elapsed time was being incremented multiple times per frame. I can't remember if this was because of the while loop I introduced to support compositing effects, or if it happened after I ran multiple effects with the same text_id via `cf_draw_push_text_id`

waldnercharles: If we're looking for something more explicit I could introduce another bool, but, this one seemed to have the same lifetime semantics that I needed.

PR #520, opened 2026-07-20
Replace vsync/vsync_mailbox setters with a unified present mode API

pusewicz:
Fixes `cf_app_set_vsync_mailbox(true)` silently no-op'ing on Metal: SDL's Metal driver rejects mailbox present mode, `SDL_SetGPUSwapchainParameters` returns `false`, but the old code discarded that return value and unconditionally cached `app->vsync = true` — so the app believed mailbox was active when the swapchain never actually changed.

`SDL_GPUPresentMode` is a plain enum, not a bitmask — only one mode is ever active on the swapchain. The old `cf_app_set_vsync`/`cf_app_set_vsync_mailbox` pair modeled this as two independent booleans over one shared state, which couldn't distinguish vsync from mailbox, couldn't represent a rejected request, and could silently clobber each other's result.

This replaces them with a single `cf_app_set_present_mode(CF_PresentMode)` / `cf_app_get_present_mode()` pair mirroring SDL's own model. The setter only commits the new mode when the backend actually applies it, so a rejected mailbox request on Metal now returns `false` and leaves the previous mode untouched instead of lying.

**Breaking change:** `cf_app_set_vsync`, `cf_app_set_vsync_mailbox`, and `cf_app_get_vsync` are removed rather than deprecated — this codebase has no compat-shim convention, and three overlapping boolean setters over one underlying state wasn't worth keeping around.

**⚠️ CI dependency on #517:** `test/test_app.cpp` needs a real swapchain and isn't gated behind `#if 0`, so Linux CI is expected to fail until #517 (headless GPU context via Xvfb + lavapipe) merges first. macOS/Windows CI should be unaffected.

### DIFF
--- a/include/cute_app.h
+++ b/include/cute_app.h
@@ -656,6 +656,45 @@ CF_INLINE const char* cf_msaa_string(CF_MSAA msaa) {
 }
 
 /**
+ * @enum     CF_PresentMode
+ * @category app
+ * @brief    Describes a swapchain present mode -- vsync off, plain vsync, or low-latency mailbox vsync.
+ * @related  cf_app_get_present_mode cf_app_set_present_mode cf_present_mode_string
+ */
+#define CF_PRESENT_MODE_DEFS \
+	/* @entry Vsync is off; frames present immediately (tearing may occur). */ \
+	CF_ENUM(PRESENT_MODE_IMMEDIATE, 0) \
+	/* @entry Vsync is on; frames present synchronized to the display refresh. */ \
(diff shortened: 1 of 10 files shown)


pusewicz:
Note: `test/test_app.cpp`'s regression tests are enabled (not gated) since they need a real `make_app()`/swapchain, same as `test_markups.cpp`. That means Linux CI on this PR will fail until #517 (headless GPU context via Xvfb + lavapipe) merges first -- macOS/Windows should be unaffected since they run the test binary natively. Happy to rebase once #517 lands.

PR #521, opened 2026-07-20
Recycle draw-command sprite buffers to eliminate per-frame churn

pusewicz:
**Depends on #517** - the new test needs a real graphics context (same as `test_markups`), so CI won't pass until #517 lands.

## Summary

Every frame, CF built a fresh little buffer for each draw command's sprite data, then threw it away and rebuilt it from scratch next frame - even when nothing changed. This recycles those buffers instead of freeing and reallocating them every frame. No API or behavior change, just less allocator churn (most noticeable in scenes with many draw commands per frame).

## Test plan
- [x] Full test suite passes, including a new regression test (`test_draw_recycle.cpp`) proving steady-state alloc/free churn is zero
- [x] Verified clean under ASan/UBSan
- [x] Verified empirically: ~200 allocs+frees/frame before this change, 0/0 after
- [ ] CI blocked on #517

https://claude.ai/code/session_013Vk9Mr4TGYFZ8n8vef6P17

### DIFF
--- a/src/cute_app.cpp
+++ b/src/cute_app.cpp
@@ -592,6 +592,7 @@ int cf_app_draw_onto_screen(bool clear)
 	s_draw->shaders.set_count(1);
 	s_draw->verts.clear();
 	s_draw->draw_item_order = 0;
+	for (int i = 0; i < s_draw->cmds.count(); ++i) s_draw->recycle_items(s_draw->cmds[i]);
 	s_draw->cmds.clear();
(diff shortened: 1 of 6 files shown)


RandyGaul:
Hmm this is probably not going to affect the bottom-line of any performance situation in practice, but, the change is simple enough I think it's fine

RandyGaul:
Likely going to punt this one in favor of the new tile based renderer

PR #524, opened 2026-07-21
Tile based renderer for Draw API

RandyGaul:
Rewrites `cute_draw` as a command renderer. Public API for drawing unchanged; the CPU records compact commands and the GPU does the rest.

**Why?** The old renderer streamed a mesh: the CPU fit a tight quad around every shape and expanded it into six fat vertices (~900 bytes per shape, rebuilt and re-uploaded every frame), and the GPU blended each overlapping quad through the framebuffer one at a time. The CPU spent its time describing *coverage*; the GPU spent its bandwidth on *overdraw*.

**The idea:** shapes are already SDFs in the fragment shader, so ship commands instead of geometry. Each drawable is one 80-byte command plus a small shared payload. Two GPU paths consume the exact same upload:

- **Instanced** -- one instance per command; the vertex shader derives the coverage quad from shape params (the old CPU-side quad fitting, moved on-GPU), the fragment shader evaluates the SDF.
- **Tiled** -- compute bins commands into 16px-tile lists with SDF-exact culling, then a fullscreen walk composites each pixel's list in registers: one framebuffer write per pixel, and opaque-cover culling drops everything hidden beneath an opaque shape without shading it.

Per-batch heuristics route to whichever wins (raster at moderate overdraw, tile walk when opaque stacking makes culling profitable). Both preserve paint order and are pixel-diffed against each other across the test suite. Recording got simpler too: no tight-fit quads or triangulation (polylines dropped ~200 lines of joint case analysis for bisector-clipped capsule SDFs that partition strokes exactly).

**New features that fell out of the design:**

- `cf_make_custom_shape` -- register a GLSL SDF snippet as a first-class shape. Batches with everything, and AA, stroke/fill, tile culling, and occlusion come for free. The snippet compiles into all shape pipelines, including the binning compute.
- `cf_draw_shape_group_begin/op/end` -- boolean ops (union/subtract/intersect, optional smooth blending) over plain draw calls, composited as **one** shape: translucent composites blend once, and the boolean result can be outlined as a single continuous stroke. Data-driven, works everywhere including web.
- `cf_draw_arrow` is now a single SDF (shaft unioned with head -- no more double-blend at the seam).
- GLES3/WebGL2 runs the same command renderer: storage buffers are emulated as texel fetches, instanced path only (compute-less). All draw tests pass on the GLES backend.

**Perf** (RTX 5070, 1280x720, vsync off, avg of 200 frames; the 6000-object scene records ~10k commands once polyline segments are counted):

| Scene | Old mesh | This branch |
| --- | --- | --- |
| 6000 translucent mixed shapes | 13.96 ms | 1.99 ms |
| ...CPU record portion | ~2.0 ms | 1.30 ms |
| 2000 large stacked opaque shapes | 6.6+ ms | 0.46 ms |
| GLES3 backend, same 6000-shape scene | -- | 2.32 ms |

**Breaking:**

- `CF_Vertex` / `CF_VertexFn` / `cf_set_vertex_callback` removed. Custom SDF shapes and `shader()` stubs are the replacement.
- `cute_spritebatch.h` renamed to `cute_atlas_cache.h` (it is a pure runtime atlas cache now; API renamed to `atlas_cache_*`).
- Precompiled user draw shaders must be regenerated with the new `cute-shaderc`.
- Windows/D3D12 builds must ship `dxcompiler.dll` / `dxil.dll` next to the exe (storage-buffer shaders require DXIL).

https://github.com/user-attachments/assets/9f7a9761-79c5-44e2-9a1b-bde570d9c597


RandyGaul:
Both Codex findings were valid and are fixed in 543f1e81:

- **Polyline segment loss (P2, actually P1-worthy)**: the geometry push sat outside the segment loop, so 3+ point polylines (and beziers) rendered only their final segment. The path-vs-path diff tests were blind to it since both paths drew the same wrong thing. Fixed by pushing one geometry per segment; a new probe test covers first/middle/last segments explicitly.
- **Cross-texture paint order (P2)**: the first atlas callback rendered the whole stream and later textures' sprites appended on top. Atlas callbacks now only stash per-sprite uvs into a per-flush table, and rendering happens after the flush in paint order, splitting draws at texture changes (this also deleted the interleave merge machinery). The multi-atlas test gains a glyphs-under-a-shape probe that fails on the old ordering.

PR perf table updated: the polyline fix means benches now record every segment (~10k commands for the 6000-object scene), so honest numbers are 1.99 ms translucent / 0.46 ms opaque-tiled vs the old mesh renderer's 13.96 / 6.6+ ms. Suite at 195/195.

PR #552, opened 2026-08-05
3D shippability shore-up: D3D12 depth-cube warning, cm_interleave, Shipping 3D docs

RandyGaul:
Follow-up to the 3D API shippability review. Three pieces:

**D3D12 depth-cube guard** -- SDL_GPU's D3D12 driver creates only 2D depth views, so rendering into a depth cube face silently produces nothing, and D3D12 is the default driver on Windows. Canvas creation now prints a loud warning pointing at the portable patterns (six 2D depth canvases, a depth atlas, or the `point_light` sample's color-encoded distance cube).

**`cm_interleave`** -- the streams-to-vertex-buffer glue everyone was hand-rolling. One `CM_VertexAttribute { stream, offset }` per output attribute, documented defaults for missing streams (rigid-skinning weights, opaque white color, identity-ish tangent) so one attribute list serves skinned and unskinned meshes. `model3d` now uses it; `test_model_interleave` pins real, generated, and defaulted streams.

**Shipping 3D docs page** -- consolidates every per-backend caveat (D3D12 depth cubes, the web tier's capability list, MSAA/MRT exclusivity, depth-sampling requirements, no GPU timing) and the pre-flight recipes: BCn textures with mips via DDS + mipped-PNG web fallback, chunked draw-list frustum culling, and the AA stance. Wired into the nav and linked from 3D Drawing. Also updates the 2D-only positioning in the README, docs front page, getting started, mkdocs description, and CMake description to '2D and 3D'.

Verified: `test_model` (3/3, incl. new case) and `test_draw3d` (22/22) pass; `model3d` renders correctly through the new interleave path (screenshot harness).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 4.2)
 project(cute_framework
-	DESCRIPTION "The cutest framework out there for creating 2D games in C++!"
+	DESCRIPTION "The cutest framework out there for creating 2D and 3D games in C++!"
 	HOMEPAGE_URL "https://github.com/RandyGaul/cute_framework"
 	LANGUAGES C CXX
 	VERSION 1.1.1
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 <img alt="Discord" src="https://img.shields.io/discord/432009046833233930?label=discord">
 </p>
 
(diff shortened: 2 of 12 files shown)


RandyGaul:
Consolidated into #555 -- Randy wants one PR for this session's work. All commits from this branch are merged there.

PR #575, opened 2026-08-10
Fix HiDPI half-size regression: default 2d projection is logical points

pusewicz:
3f5a8283 wired `cf_draw_on_app_canvas_resized` to the canvas's *pixel* dimensions. The default canvas is `window_points * pixel_scale`, so on a 2x display the projection extent doubled and everything drawn in point space rendered at half size — from the second frame on, since the hook never refreshed `mvp` and `reset_cam` only latches it at end of frame. At 1x displays points and pixels coincide, which is why CI never saw it, and why it's invisible on most dev machines too.

`s_canvas` now hands the hook window points, and the hook refreshes `mvp` (mirroring `cf_draw_projection`) and the AA factor. When a refresh lands mid draw-list-recording or inside a push/pop pair it parks in `pending_projection` and `reset_cam` applies it at the frame boundary — applying immediately would bake the ortho into recorded draw lists (doubled again at replay) or be clobbered by the pop.

DPI bugs were untestable on 1x machines, so `cf_app_force_pixel_scale` (internal) pins `pixel_scale` through the real recreation path: the new `test_hidpi` suite readback-verifies the projection contract, the resize frame, one-shot `cf_app_set_canvas_size` behavior, and draw-list recording across a resize, at a forced 2x on any machine.

On this Retina Mac, master currently fails 20 `test_draw_tiled` cases; this branch is 346/346.

First of a 3-PR series (viewport/scissor scaling and 3d stroke thickness build on this one's test infra) — holding the rest until this lands so each diff stays reviewable on its own.

### DIFF
--- a/src/cute_draw.cpp
+++ b/src/cute_draw.cpp
@@ -1001,8 +1001,21 @@ static void s_init_atlas_cache(int w, int h)
 	}
 }
 
+// The default 2d projection spans the window's LOGICAL size (points, never canvas pixels --
+// on a 2x display the canvas is twice as large, and a pixel-sized ortho would halve every
+// draw in point space).
+static CF_M3x2 s_default_projection()
+{
+	return ortho_2d(0, 0, (float)app->w, (float)app->h);
+}
+
 void CF_Draw::reset_cam()
 {
+	// Apply a refresh parked by cf_draw_on_app_canvas_resized.
+	if (pending_projection && !recording_list && projection_stack.count() == 0) {
+		projection = s_default_projection();
+		pending_projection = false;
+	}
 	cam_stack.clear();
 	cam_stack.add(cf_make_identity());
 	mvp = projection;
@@ -1025,7 +1038,7 @@ void cf_make_draw()
 {
 	s_draw = CF_NEW(CF_Draw);
 	s_draw->path_image_id_gen = CF_PATH_ID_RANGE_LO;
-	s_draw->projection = ortho_2d(0, 0, (float)app->w, (float)app->h);
+	s_draw->projection = s_default_projection();
 	s_draw->reset_cam();
 	s_draw->uniform_arena = cf_make_arena(32, CF_MB);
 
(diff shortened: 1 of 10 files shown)


src/cute_draw.cpp
```diff
@@ -5198,11 +5204,34 @@ static void s_process_command(CF_Canvas canvas, CF_Command* cmd, CF_Command* nex
...
+		s_draw->pending_projection_h = h;
+		return;
+	}
+	s_draw->pending_projection = false;
+	s_draw->projection = ortho_2d(0, 0, (float)w, (float)h);
+	// Refresh mvp too, mirroring cf_draw_projection: reset_cam only re-latches it at end of
+	// frame, so without this the rest of the resize frame draws with the stale matrix. The
+	// AA factor divides by pixel_scale, so it is equally stale on a density change.
+	CF_MUL_M32_M32(s_draw->mvp, s_draw->projection, s_draw->cam_stack.last());
+	s_draw->set_aaf();
```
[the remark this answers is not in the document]
pusewicz: Fixed in ecca88b1: clamps only the value fed to `ortho_2d` (via a local `ortho_w/ortho_h`), so `w/h` still report the real logical size to callers and `CF_ASSERT(w == app->w && h == app->h)` is untouched.

src/cute_draw.cpp
```diff
@@ -5249,13 +5263,22 @@ static void s_process_command(CF_Canvas canvas, CF_Command* cmd, CF_Command* nex
...
-	// startup and never again, so any resize (cf_app_set_size or a user dragging a resizable
-	// window) silently rescaled every world-space 2d draw. Refresh it with the canvas; a
-	// custom cf_draw_projection is per-frame state and simply overrides this as usual.
-	if (s_draw) s_draw->projection = ortho_2d(0, 0, (float)w, (float)h);
+void cf_draw_on_app_canvas_resized()
+{
+	if (!s_draw) return;
+	// Applying mid draw-list-recording would stomp the recording's identity space, and
+	// inside a cf_draw_push/pop pair the pop would just clobber it. Park the refresh;
+	// reset_cam applies it at the frame boundary.
```
bullno1, line 5339: When would this happen? The event loop drains all event before user code can run so a resize event can't happen during draw.

src/cute_draw.cpp
```diff
@@ -5249,13 +5263,22 @@ static void s_process_command(CF_Canvas canvas, CF_Command* cmd, CF_Command* nex
...
+	if (s_draw->recording_list || s_draw->projection_stack.count() > 0) {
+		s_draw->pending_projection = true;
+		return;
+	}
+	s_draw->pending_projection = false;
+	// cf_draw_projection refreshes mvp too: reset_cam only re-latches it at end of frame,
+	// so the rest of the resize frame would otherwise draw with the stale matrix. Ditto
+	// the AA factor, which divides by pixel_scale.
+	cf_draw_projection(s_default_projection());
```
bullno1, line 5348: This is probably correct most of the time but it overrides what was set by `cf_draw_projection` which was supposed to be sticky.

bullno1 (commented):
`cf_draw_projection` is exposed to the user but it conflicts with the automagic behaviour. And the automagic way also forces one to provide an override for testing on non-HiDPI monitor. 

I think there should be no automagic behaviour at all:

* Expose dpi scale: `cf_app_get_dpi_scale`
* Expose dpi change event: `cf_app_dpi_scale_was_changed`
* Have `cf_app_{get,set}_pixel_scale`. This is now a user controlled value just like window and canvas size. Text and shapes can reference this.
* Depending on the creation bit, create the canvas, set pixel scale and projection only **once**. Maybe even get rid of the bit and just be aware by default. Most of the time, the window is not resizable and DPI does not change. To handle it, copy from the sample or doc.

Then we can update the dpi sample and document on how to handle the events (resize, dpi change...) and how to resize the canvas and recalculate projection. The sample can also force pixel scale to facilitate testing on non-HiDPI screen. There is no more "override" since pixel scale is a user controlled variable, just that it has an initial default. 

The sample can switch between:

• Scale = forced 1x
• Scale = natural (reported by SDL)
• Scale = forced 2x, forced 4x...

This means handling dpi change is done the same way as resize: on the event, make the calls to the appropriate functions. Just copy and paste from samples or doc if needed. 

Basically:

* To be unaware: set pixel scale to 1, don't react to dpi event, only to resize event. Update canvas and projection accordingly.
* To be aware, set pixel scale to dpi scale, also react to dpi change event.

The one thing I haven't looked at is how window size and mouse position is reported. IIRC, SDL reports raw values from the OS which is weird. The OSX way makes the most sense. Windows is weird. Linux depends on whether it's Xorg (Windows way) or Wayland (OSX way). That part should be normalized.

bullno1:
Hmm, the two cases (aware vs unaware) are similar enough that maybe a single: `void cf_app_update_display_metrics(width, height, pixel_scale)` helper can work. It will do all: projection, canvas, pixel scale.

The user is responsible for calling it at the correct time. `pixel_scale` can be an arbitrary value for testing or just pass `cf_app_get_dpi_scale` through.

PR #579, opened 2026-08-14
Explicit user-controlled pixel scale and HiDPI fix

pusewicz:
Working demo of @bullno1's proposal from the #575 review thread (including the follow-up naming refinements): remove all event-driven HiDPI behavior and make pixel scale a plain user-controlled value.

- Startup is the only automatic step: canvas at `window_size * display_scale`, default 2d projection from the logical window size, both set once. Resizes and scale changes only raise flags (`cf_app_was_resized` / `cf_app_display_scale_was_changed`).
- Terminology follows the SDL functions underneath, per the thread. `cf_app_get_display_scale` (+`_was_changed`) is `SDL_GetWindowDisplayScale` — what the OS wants, consistent across platforms; the `dpi` names are gone. `cf_app_set_pixel_scale` is how fonts/shapes scale — arbitrary, or following the reported display scale; it does not touch the canvas (`cf_app_set_canvas_size` is separate). `cf_app_update_display(scale)` is the all-in-one helper bringing canvas, projection, and pixel scale together, usable on both resize and scale change — the w/h params are omitted since the window size is always current by the time user code reacts.
- Window size and mouse coordinates are normalized to logical points at the SDL boundary via `SDL_GetDisplayContentScale` (window creation, `cf_app_set_size`, resize events, mouse events; touch derives from window size). This is a no-op on macOS (content scale 1.0) and needs verification on Windows/X11 at >100% scaling — I can only test on a Mac.
- `cf_draw_projection` becomes truly sticky and `cf_app_set_canvas_size` persistent. The half-size regression #575 fixes dies structurally — no code path can rebuild the projection in pixel units — and all of #575's deferral/guard machinery becomes unnecessary. 351/351 tests on a 2x display (master: 319/339, the 20 failures being that regression).
- The hidpi sample carries the recipe (N follows the display scale, 1/2/4 force scales — forcing a value is also how you test HiDPI on a normal monitor); test_hidpi covers the new contracts.

Cost, stated plainly: breaking change. 44 of 76 samples use resizable windows and each needs the two-line recipe (deliberately not yet updated); canvas_modes.c and the hidpi / application_window topic docs need a rewrite (deferred together with the samples). `NO_HIGH_DPI_BIT` is kept: it is the only handle on the creation-time 1x swapchain (memory + final-blit bandwidth), which `cf_app_update_display(1)` cannot replicate.

## Before

Screenshot 2026-08-24 at 23 59 28

## After

Screenshot 2026-08-24 at 23 59 38

https://claude.ai/code/session_01LHq3Cx7xcjAQhji9m1ADJn

### DIFF
--- a/include/cute_app.h
+++ b/include/cute_app.h
@@ -221,8 +221,8 @@ typedef enum CF_AppOptionFlagBits
  * @param    display_index The index of the display to spawn upon. Set this to zero for the primary display. See `cf_get_display_list`.
  * @param    x             The x position of the window.
  * @param    y             The y position of the window.
- * @param    w             The width of the window in pixels.
- * @param    h             The height of the window in pixels.
+ * @param    w             The width of the window in logical points.
+ * @param    h             The height of the window in logical points.
  * @param    options       0 by default; a bitmask of `app_options` flags.
  * @param    argv0         The first argument passed to your main function in the `argv` parameter.
  * @return   Returns any errors on failure as a `CF_Result`.
@@ -248,6 +248,8 @@ typedef enum CF_AppOptionFlagBits
  *     }
  * @remarks  The options parameter is an enum from `app_options`. Different options can be OR'd together.
  *           Parameters `w` and `h` are ignored if the window is initialized to fullscreen mode with `CF_APP_OPTIONS_FULLSCREEN_BIT`.
+ *           Like `cf_app_set_size`, `w` and `h` are logical points: on a display scaled above 100% (e.g. Windows at 150%)
+ *           the physical window is correspondingly larger; see the hidpi sample.
  * @related  CF_AppOptionFlagBits cf_app_is_running cf_app_signal_shutdown cf_destroy_app
  */
 CF_API CF_Result CF_CALL cf_make_app(const char* window_title, CF_DisplayID display_id, int x, int y, int w, int h, CF_AppOptionFlags options, const char* argv0);
--- a/include/cute_app.h
+++ b/include/cute_app.h
@@ -467,6 +467,21 @@ CF_API bool CF_CALL cf_app_pixel_scale_was_changed(void);
 CF_API float CF_CALL cf_app_get_natural_pixel_scale(void);
 
 /**
+ * @function cf_app_apply_pixel_scale
+ * @category app
+ * @brief    Applies a pixel scale in one call: sets the scale, resizes the app canvas to window size times scale, and rebuilds the default 2d projection from the logical window size.
+ * @param    scale      The pixel scale to apply. Must be greater than zero; other values are ignored.
+ * @remarks  This is the standard reaction to a window resize or a display density change, packaged up -- equivalent
+ *           to `cf_app_set_pixel_scale`, then `cf_app_set_canvas_size(window_w * scale, window_h * scale)`, then
+ *           `cf_draw_projection` spanning the logical window size. Call it when `cf_app_was_resized` fires (passing
+ *           `cf_app_get_pixel_scale` to keep the current scale), when `cf_app_dpi_scale_was_changed` fires (passing
+ *           `cf_app_get_natural_pixel_scale` to track the display), or any time with an arbitrary scale for testing.
+ *           It overwrites a custom `cf_draw_projection` -- re-apply yours after, if you use one. See the hidpi sample.
+ * @related  cf_app_set_pixel_scale cf_app_get_pixel_scale cf_app_get_natural_pixel_scale cf_app_set_canvas_size cf_app_dpi_scale_was_changed
+ */
+CF_API void CF_CALL cf_app_apply_pixel_scale(float scale);
+
+/**
  * @function cf_app_set_size
  * @category app
  * @brief    Sets the size of the window in logical points.
@@ -1015,6 +1030,7 @@ CF_INLINE float app_get_pixel_scale() { return cf_app_get_pixel_scale(); }
 CF_INLINE void app_set_pixel_scale(float scale) { cf_app_set_pixel_scale(scale); }
 CF_INLINE bool app_pixel_scale_was_changed() { return cf_app_pixel_scale_was_changed(); }
 CF_INLINE float app_get_natural_pixel_scale() { return cf_app_get_natural_pixel_scale(); }
+CF_INLINE void app_apply_pixel_scale(float scale) { cf_app_apply_pixel_scale(scale); }
 CF_INLINE void app_center_window() { cf_app_center_window(); }
 CF_INLINE bool app_was_resized() { return cf_app_was_resized(); }
 CF_INLINE bool app_was_moved() { return cf_app_was_moved(); }
--- a/include/cute_app.h
+++ b/include/cute_app.h
@@ -427,7 +427,7 @@ CF_API bool CF_CALL cf_app_display_scale_was_changed(void);
  *           user-controlled value, like the window or canvas size: it starts at the display scale the OS wants
  *           (e.g. 2.0f on a 2x Retina display, or 1.0f if `CF_APP_OPTIONS_NO_HIGH_DPI_BIT` was passed to
  *           `cf_make_app`) and afterwards changes only through `cf_app_set_pixel_scale` or `cf_app_update_display`.
- * @related  cf_app_set_pixel_scale cf_app_update_display cf_app_pixel_scale_was_changed cf_app_get_display_scale cf_app_get_size cf_app_get_canvas_width cf_app_get_canvas_height
+ * @related  cf_app_set_pixel_scale cf_app_update_display cf_app_get_display_scale cf_app_get_size cf_app_get_canvas_width cf_app_get_canvas_height
  */
 CF_API float CF_CALL cf_app_get_pixel_scale(void);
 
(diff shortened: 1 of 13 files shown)


docs/topics/hidpi.md
```diff
@@ -1,33 +1,53 @@
...
 
-Modern displays (Apple Retina, many 4K laptops, phones) pack several physical pixels into a single logical "point". Cute Framework handles this for you: your game works in **logical points**, while CF renders internally at the display's **physical pixel** resolution. You get crisp output on high-density displays without changing any of your drawing code.
+Modern displays (Apple Retina, many 4K laptops, phones) pack several physical pixels into a single logical "point". Cute Framework's drawing API works in **logical points**, while rendering happens into a canvas sized in **physical pixels**. How many physical pixels one point maps to is the **pixel scale** — a plain user-controlled value, just like the window size or the canvas size.
+
+At startup CF does the right thing once: it creates the app canvas at `window_size * display_scale` and sets the default 2d projection from the logical window size. After that, nothing changes behind your back. A fixed-size window on a single display — most games — needs no DPI code at all. A resizable window, or one that migrates between displays of different density, reacts to those events with one call (see the recipe below).
```
bullno1: Maybe define the terms, esp `window_size` because "size" is very loaded in this context.

samples/hidpi.c
```diff
@@ -1,27 +1,35 @@
 /*
-	hidpi.c -- HiDPI / Retina rendering visual verification.
+	hidpi.c -- HiDPI / Retina rendering, the manual way.
```
bullno1, line 2: I think it should prints out the mouse coordinate too to see if mouse coordinate is translated correctly.

Anw, it's correct.

docs/topics/hidpi.md
```diff
@@ -1,33 +1,53 @@
...
+
+	// ... draw and present as usual ...
+}
+```
 
-When you need to convert between points and physical pixels — for example, sizing an offscreen canvas to match the swapchain — use `cf_app_get_pixel_scale`.
+Drop the second block to keep a fixed scale regardless of display. If you drive a custom projection (your own [`cf_draw_projection`](../draw/function/cf_draw_projection.md)), re-apply it after the helper — `cf_app_update_display` writes the default one, and it is the *only* thing in CF that ever writes a projection on your behalf after startup.
 
-## How the default canvas resizes
+Because the pixel scale is just a value, forcing one is also how you test HiDPI behavior on a normal monitor: `cf_app_update_display(2.0f)` renders exactly like a 2x display, no Retina hardware required. [`cf_app_pixel_scale_was_changed`](../app/function/cf_app_pixel_scale_was_changed.md) reports true for one frame after any change, for decoupled systems that want to notice.
```
bullno1: I don't really see the point of this but it's probably just me.

The change originates from the user anyway so why an event?

pusewicz: @bullno1 Why a `cf_app_pixel_scale_was_changed` event? Or why an event after `cf_app_update_display` gets called? If the latter, is so that the change can be handled by the same code as the one that would receive such an event in the event of moving a window from a HiDPI screen to a standard DPI screen, or by changing the resolution in the OS from a HiDPI one to a standard DPI resolution (which, is the case for example in macOS).

samples/hidpi.c
pusewicz: Done in https://github.com/RandyGaul/cute_framework/pull/579/commits/2b7380751ba9509a006163c4f8c310051087dd61

docs/topics/hidpi.md
pusewicz: Done in https://github.com/RandyGaul/cute_framework/pull/579/commits/0e87b5c5260bfc7be01ef792a18fe89f4312be3a

docs/topics/hidpi.md
bullno1: Why `cf_app_pixel_scale_was_changed` event. Because it only happens after calling `cf_app_set_pixel_scale`.

pusewicz: @bullno1 You are right. Got rid of it.

pusewicz:
@RandyGaul @bullno1 I trimmed the PR. Please have a look and let me know what needs clarifying.

src/cute_input.cpp
```diff
@@ -625,32 +623,36 @@ void cf_pump_input_msgs()
...
+			float cs = cf_app_get_content_scale();
+			app->mouse.x = event.button.x / cs;
+			app->mouse.y = event.button.y / cs;
 			if (event.button.clicks == 1) {
 				app->mouse.click_type = CF_MOUSE_CLICK_SINGLE;
 			} else if (event.button.clicks == 2) {
 				app->mouse.click_type = CF_MOUSE_CLICK_DOUBLE;
 			}
-			break;
+                }	break;
```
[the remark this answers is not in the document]
pusewicz: Fixed in https://github.com/RandyGaul/cute_framework/pull/579/commits/a17f13dd38e8dbbf86e05f8b38fde56f0eb40951.

include/cute_app.h
```diff
@@ -400,43 +400,75 @@ CF_API void CF_CALL cf_app_show_window(void);
...
- * @param    w          The width of the window in pixels.
- * @param    h          The height of the window in pixels.
- * @related  cf_app_get_size cf_app_get_position cf_app_set_position
+ * @brief    Sets the size of the window in logical points.
+ * @param    w          The width of the window in logical points.
+ * @param    h          The height of the window in logical points.
+ * @remarks  Only the window changes. The app canvas and the default 2d projection keep their current size --
+ *           update them alongside if desired, e.g. `cf_app_set_canvas_size` and `cf_draw_projection`; see the
+ *           hidpi sample for the recipe.
+ * @related  cf_app_get_size cf_app_get_position cf_app_set_position cf_app_set_canvas_size
```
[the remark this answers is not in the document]
pusewicz: Fixed in https://github.com/RandyGaul/cute_framework/pull/579/commits/0f68acf6290fc53be47528bdb2fcbb9ff8b68389 -- `w`/`h` documented as logical points, matching `cf_app_set_size`.

test/test_hidpi.cpp
```diff
@@ -0,0 +1,392 @@
+/*
+	Cute Framework
+	Copyright (C) 2026 Randy Gaul https://randygaul.github.io/
```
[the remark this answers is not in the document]
pusewicz: 2026 is already the year in 8 other files here (e.g. test/test_app_shared.cpp, test/test_alloc.cpp), so leaving it as is.

include/cute_app.h
```diff
@@ -729,11 +761,11 @@ CF_API CF_Canvas CF_CALL cf_app_get_canvas(void);
...
- *           This is a one-shot override. The app's canvas is automatically recreated at window size (in points) times
- *           `cf_app_get_pixel_scale` on every canvas recreation event -- a window resize, moving to a display with a
- *           different pixel density, `cf_app_set_size`, or `cf_app_set_msaa` -- so a custom size lasts only until the
- *           next such event. For a persistent fixed-resolution render target (e.g. a retro/pixel-art look) make your
- *           own canvas with `cf_make_canvas` and draw it scaled-up with `cf_draw_canvas`; see the canvas_modes sample.
+ *           The canvas keeps this size until the next `cf_app_set_canvas_size` call -- nothing resizes it behind
+ *           your back. It is created once at startup at window size (in points) times the display's pixel density;
+ *           after that, window resizes and display density changes only raise `cf_app_was_resized` /
+ *           `cf_app_display_scale_was_changed`, and resizing the canvas in response is up to you (see the hidpi sample
+ *           for the recipe).
```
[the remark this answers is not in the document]
pusewicz: Header contracts fixed in https://github.com/RandyGaul/cute_framework/pull/579/commits/60d85155a9ad790971798398e577ba5c5ee5a5df (`cf_draw_projection`) and https://github.com/RandyGaul/cute_framework/pull/579/commits/0f68acf6290fc53be47528bdb2fcbb9ff8b68389 (`cf_make_app`). The topic docs were reverted from this PR deliberately; they get rewritten together with the sample updates, listed under the cost paragraph in the description.

src/cute_app.cpp
```diff
@@ -305,10 +297,12 @@ CF_Result cf_make_app(const char* window_title, CF_DisplayID display_id, int x,
...
 		if (options & CF_APP_OPTIONS_HIDDEN_BIT) flags |= (SDL_WINDOW_HIDDEN | SDL_WINDOW_MINIMIZED);
 
+		float creation_content_scale = SDL_GetDisplayContentScale(display_id ? display_id : SDL_GetPrimaryDisplay());
+		if (creation_content_scale <= 0) creation_content_scale = 1.0f;
 		SDL_PropertiesID props = SDL_CreateProperties();
 		SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_TITLE_STRING, window_title);
-		SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_WIDTH_NUMBER, w);
-		SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, h);
+		SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_WIDTH_NUMBER, (int)CF_ROUNDF(w * creation_content_scale));
+		SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, (int)CF_ROUNDF(h * creation_content_scale));
```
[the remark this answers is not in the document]
pusewicz: Fixed in https://github.com/RandyGaul/cute_framework/pull/579/commits/0f68acf6290fc53be47528bdb2fcbb9ff8b68389 -- `cf_make_app` now documents `w`/`h` as logical points, matching `cf_app_set_size`.

test/test_hidpi.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+// NO_HIGH_DPI only pins the INITIAL scale at 1.0 (the window gets a 1x backbuffer, so the
+// display scale is 1.0 too). The value stays user-controllable afterwards.
+TEST_CASE(test_hidpi_no_high_dpi_initial_scale)
+{
+	if (!test_make_app(LOGICAL_W, LOGICAL_H, CF_APP_OPTIONS_NO_HIGH_DPI_BIT)) return true; // Headless CI: no display/GPU.
+	HidpiGuard guard;
+
+	REQUIRE(cf_app_get_pixel_scale() == 1.0f);
+	REQUIRE(cf_app_get_display_scale() == 1.0f);
```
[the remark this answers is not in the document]
pusewicz: Fixed in https://github.com/RandyGaul/cute_framework/pull/579/commits/539b7181117215b2d877bc624f30c594b02d634b -- `NO_HIGH_DPI_BIT` only pins the pixel scale; the display scale stays whatever SDL reports, and the test no longer asserts it is 1.0.

RandyGaul:
Thanks for digging into this and for #575's diagnosis -- the half-size regression was real. Landing the fix as #607 instead: the hook now rebuilds the projection from the window size in points (as startup already did), with a forced-2x test so 1x CI catches it.

Closing this one because I don't want to take the rest of the model change with it. CF's contract stays: window sizes in points, canvas and projection automatic on resize, pixel scale read-only. I'd rather not make resize handling the user's job, and I don't want to import SDL's content-scale semantics on Windows/X11 (fractional scales, physical windows growing with the desktop slider, and the NO_HIGH_DPI blit becoming a 1.5x nearest upscale). If honoring the desktop slider ever comes up, the shape would be a single opt-in flag on cf_make_app, not setters.

#582 (the mouse readout) is merged.

PR #581, opened 2026-08-17
Rename dpi_scale to display_scale

pusewicz:
`cf_app_get_dpi_scale` / `cf_app_dpi_scale_was_changed` read like the OS's vague "suggested UI content scale" concept. The value is actually `SDL_GetWindowDisplayScale` -- the OS's points-to-pixels conversion for the window's display -- so this renames it to say that.

**Breaking change:** hard rename, no compat alias.

Also drops `CF_App::dpi_scale_prev`, which was written once at init and never read.

First slice of the #579 split -- pure rename, zero behavior change.


pusewicz:
@RandyGaul Is there anything holding this PR?

RandyGaul:
@pusewicz nope

PR #582, opened 2026-08-17
samples/hidpi: add live mouse coordinate readout

pusewicz:
Adds a `mouse: screen (x, y) -> world (x, y)` readout line to the hidpi sample, next to the existing `pixel_scale` line -- so HiDPI mouse-coordinate mapping can be eyeballed against known shape positions at any pixel scale, not just glyph/shape crispness.

Second slice of the #579 split -- pure sample addition, no engine changes.

**Note**: The scale of the rendered content is incorrect, which will be fixed in the upcoming PRs.

hidpi_after


pusewicz:
@RandyGaul this adds extra info about the mouse coords to the hide sample.

PR #608, opened 2026-09-11
Make layers first class with their own command queues

bullno1:
Layers are now first class instead of a command attribute.
Each layer has its own command queue. Switching layer with `cf_push_layer` and `cf_pop_layer` just changes the layer being submitted to.
Layers not used for an entire frame will be pruned.

Compared to the previous method:

* `CF_DRAW_SPLIT_3D_LAYER` is no longer possible. Layers can be used without breaking batches.
  Layers can now be used for multipass drawing: Iterate over an object list, submit draw for each pass in a separate layer and the commands should still be merged.
* Repeated switching would not grow the queue. Previously every layer switch creates a new command.
* No sorting or rather less sorting. Only the small list of layers have to be sorted.

The drawbacks:

* Code that previously use generated layer id (for example y sorting in 2D game) will create a large number of layer objects.

All existing behavior should be preserved:

* Setting states (uniform, texture) in one layer and then switch to the next layer still works. Non-drawing are moved into the new layers
* Draw lists requeue their commands per layer
* Opaque/translucent partition now runs per layer
* Cross layer merging when state does not change


RandyGaul:
Reviewed and benchmarked; this is a clear win and I'm going to land it once CI is green.

I pushed two small commits onto the branch (it's maintainer-editable):

- **Export `s_draw`.** The new tests read the layer queues through it, and the Windows DLL build didn't export it (the msvc/shared job's LNK2001). Exported the same way `app` is.
- **Per-flush arena reset guard.** Moving the reset to frame end is right for a partial `cf_render_layers_to`, but a loop that renders to canvases without ever presenting would grow the arena forever. The flush now resets it whenever no queued command references it (every flush in the common case), with the frame-end reset as the fallback. Your partial-render test still passes.

On the one drawback you list -- one layer per object for y-sorting -- I measured it against master, Debug builds, best-of-N frames of quads with pseudo-random layer ids:

| quads / distinct layers | master | this PR |
|---|---|---|
| 5000 / 8 | 26.1 ms | 4.5 ms |
| 5000 / 256 | 29.4 ms | 5.2 ms |
| 5000 / 5000 | 29.7 ms | 9.1 ms |
| 20000 / 20000 | 136.9 ms | 42.3 ms |

Even the worst case is 3x faster than master, since master paid two fresh commands per push/pop pair plus a stable sort of the whole command stream. So the per-layer heap objects are a non-issue in practice.

Full suite green locally on Windows in both static and shared builds.

bullno1:
Err, I think the force push overwrote your reset guard

pusewicz:
@bullno1 `git push --force-with-lease`

pusewicz:
@RandyGaul We gotta add the reset guard again?

bullno1:
As with everything in git, the useful option is not default

RandyGaul:
What do you need me to do?

bullno1:
@RandyGaul your code for "Per-flush arena reset guard". You pushed it around the same time I force pushed on my own to fix the CI problem so it's gone.

PR #610, opened 2026-09-11
Use $CLAUDE_PROJECT_DIR for hook commands

pusewicz:
Claude Code hooks run with the tool's cwd, which isn't guaranteed to be the repo root. Using `$CLAUDE_PROJECT_DIR` makes the hook commands resolve correctly regardless of where Claude Code is invoked from.

https://claude.ai/code/session_01GfSFHpvVeNJUHWdQzBPPY2


pusewicz:
This was failing at times in Claude, when working in a worktree and executing the command verbatim. Adding the `CLAUDE_PROJECT_DIR` variable, ensures we always use an absolute path.

PR #613, opened 2026-09-13
gles: bind emulated storage textures from the top down

### LINKED ISSUE #612: GLES3/WebGL2: GL_INVALID_OPERATION (glDrawArraysInstanced, sampler location conflict) with multiple bound textures, after #611
## Summary

On the Emscripten/WebGL2 backend, a multi-pass 2D deferred renderer with several custom draw shaders throws `GL_INVALID_OPERATION: glDrawArraysInstanced: Two textures of different types use the same sampler location` on every frame, and the affected draws render nothing. A minimal single-sprite scene (the `basicsprite` sample) does not reproduce it, so this looks specific to shaders/materials that bind multiple textures together, and appeared after pulling in #611.

## Environment

- `emcc`/`emcmake` 6.0.9-git (Homebrew), macOS (Apple Silicon)
- cute_framework at `10917046` (merge of #611, "gles-ring-and-samplers"; upstream sync, no local changes)
- Browser: Chrome (headless, `--use-angle=swiftshader`) — **not yet confirmed on a non-software GL driver**; see Caveat below

## Repro

Downstream project: [pusewicz/slowhand](https://github.com/pusewicz/slowhand), a 2D game with a deferred lighting pipeline — several `R16G16B16A16_FLOAT`/`R16_FLOAT` canvases (albedo, emissive, light, sunlight, fog) composited together, plus an instanced rain simulation (`rain_sim`/`rain_drop`) that reads multiple of those canvases as textures in one draw. Building it for web (`emcmake cmake ... -DCMAKE_BUILD_TYPE=Release` + the project's own Emscripten CMake block) and loading the result in a browser:

```
GL_INVALID_OPERATION: glDrawArraysInstanced: Two textures of different types use the same sampler location.
```

repeats every frame, and the canvas shows a flat clear color instead of the scene — the composite/rain draws are being rejected.

**Discriminator:** built and ran `vendor/cute_framework/samples/basicsprite` (a single easy-sprite draw) the same way, same toolchain, same machine — zero GL errors, renders correctly. That rules out the emsdk version and the general Emscripten/GLES3 setup, and points at something specific to a draw binding several textures with (presumably) different sampler-variant requirements in one call.

**Suspect:** 9855683 ("Filter override applies to u_image only, via per-texture sampler variants", under #611) reworked how `cute_graphics_gles.cpp` caches a per-texture nearest/linear sampler variant and applies the filter override "only to the u_image binding." A material binding several textures — some via `u_image`, others as additional named samplers, or the same texture bound with two different filter overrides across draws sharing a batch — seems like the kind of case this refactor's per-texture caching could get wrong, producing two different sampler declarations resolved to the same GL sampler/texture-unit location. I haven't isolated the exact draw or texture combination yet.

## Caveat

I could only test this in a sandboxed headless Chrome using SwiftShader (software GL), and couldn't get a real-GPU cross-check: `CF_APP_OPTIONS_GFX_OPENGL_BIT` (running the same GLES3 backend natively, per `docs/topics/emscripten.md`) fails to create a window at all on native macOS (`WARNING -- OpenGL ES is not intended for non-web builds!` / `Failed to create the application window`) — no ANGLE, so that path isn't usable here as a cross-check. If someone can reproduce (or rule this out) on a real GPU/browser, or on Linux/Windows via `CF_APP_OPTIONS_GFX_OPENGL_BIT`, that would help a lot in confirming this isn't a SwiftShader-only artifact.

Happy to help narrow this down further — can share the exact shader/canvas setup, or try to build a minimal multi-texture repro in a CF sample if that's more useful than pointing at the downstream project.

pusewicz:
Fixes #612.

## Background

A GL texture unit is a numbered slot. Each sampler uniform in a linked program points at one slot. GLES 3.0 requires that two samplers of different types never point at the same slot. Chrome/Safari GL backend and rejects the draw with `GL_INVALID_OPERATION` when it happens.

CF's 2D draw pipeline emulates storage buffers as `usampler2D` textures. Before every draw it binds two of them for its own instanced vertex shader, `u_vs_storage_0` and `u_vs_storage_1`. Material textures are ordinary `sampler2D`.

## Before

Material textures were bound to units 0, 1, 2, … counting up. Emulated storage textures were hardcoded to start at unit 8 and count up. A shader with 9 textures (`u_image` plus 8 of its own) broke that assumption (#612).

| Unit | Material (9 textures)  | Storage                       | Result        |
|------|------------------------|-------------------------------|---------------|
| 0    | `u_image`              |                               | ok            |
| 1–7  | `u_tex1` … `u_tex7`    |                               | ok            |
| 8    | `u_tex8` (`sampler2D`) | `u_vs_storage_0` (`usampler2D`) | **collision** |
| 9    |                        | `u_vs_storage_1`              | ok            |

A `sampler2D` and a `usampler2D` both reference unit 8, so the browser rejects the draw:

```
GL_INVALID_OPERATION: glDrawArraysInstanced: Two textures of different types use the same sampler location.
```

With 8 material textures it worked. With 9 it silently drew nothing and raised the above error in the console.

## After

Emulated storage textures now bind from the top down, starting at `GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS - 1`, independent of the material. Vertex storage binds first, then fragment. Material textures still count up from 0. With 32 combined units (the GLES 3.0 minimum) and the worst case of 4 vertex + 4 fragment storage buffers in one draw:

| Unit | Material (9 textures) | Storage          | Result |
|------|-----------------------|------------------|--------|
| 0    | `u_image`             |                  | ok     |
| 1–8  | `u_tex1` … `u_tex8`   |                  | ok     |
| 9–23 | free                  |                  |        |
| 24   |                       | `u_fs_storage_3` | ok     |
| 25   |                       | `u_fs_storage_2` | ok     |
| 26   |                       | `u_fs_storage_1` | ok     |
| 27   |                       | `u_fs_storage_0` | ok     |
| 28   |                       | `u_vs_storage_3` | ok     |
| 29   |                       | `u_vs_storage_2` | ok     |
| 30   |                       | `u_vs_storage_1` | ok     |
| 31   |                       | `u_vs_storage_0` | ok     |

The top `CF_GLES_STORAGE_TEXTURE_RESERVE` (8) units are reserved because `cf_apply_vs_storage_buffers` and `cf_apply_fs_storage_buffers` each accept up to 4 buffers and a caller can combine 4 + 4 in one draw, as above. A draw that binds fewer (the 2D pipeline's own draws bind 2 vertex + 1 fragment, or 4 fragment) simply leaves the lower reserved units unused. A `static_assert` ties the constant to the `s_vs_storage[4]` / `s_fs_storage[4]` array sizes so it cannot drift.

We detect the number max number of texture units using `glGetIntegerv(GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS, &g_ctx.max_combined_texture_units);
` (32 is the minimum coming from the spec) and reserve 8 for VS and FS storage-buffers (4 for VS and 4 for FS).

If a material ever has enough textures to reach the reserved range, a `CF_ASSERT` in `cf_gles_apply_shader` fires instead of a silent collision.

## Repro

`samples/nine_textures.cpp` draws once with `u_image` + 8 user textures. Green window = pass. Black window + `GL_INVALID_OPERATION ... same sampler location` in the console = fail.

```sh
emcmake cmake -S . -B build-emscripten -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_SCAN_FOR_MODULES=OFF
cmake --build build-emscripten --target nine_textures
emrun build-emscripten/nine_textures.html
```

Fix confirmed in https://pusewicz.itch.io/slowhand, as it renders the whole scene correctly now.


bullno1:
How about just bind from the back? Start from  `GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS - 1` and count down? Essentially the same but it doesn't depend on how many texture the user material has. Then assert on apply material because it can't get past the limit: `GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS - 4`

pusewicz:
Done — counts down from `MAX-1` now; asserting `< MAX - 8` on the material side since it's 4 per stage (vs + fs) x 2 stages, tied to `s_vs_storage[4]`/`s_fs_storage[4]` via a static_assert so it can't drift. Thanks for the suggestion.

pusewicz:
### Code review

Found 1 issue:

1. The new regression sample's header comment claims it "only exercises the collision this guards against on the GLES3/WebGL2 backend (`CF_APP_OPTIONS_GFX_OPENGL_BIT`, or any Emscripten/web build)," but its own `make_app(...)` call never passes `CF_APP_OPTIONS_GFX_OPENGL_BIT`. On a native (non-Emscripten) build this silently runs on the default backend (unaffected by this bug), skips the GLES-specific code entirely, and still prints a "clean draw" success message — so `./gles_sampler_units` run natively gives false confidence that it validated the fix. Only an Emscripten/web build actually exercises it.

https://github.com/RandyGaul/cute_framework/blob/33d8569ea75909bb484e374e05126b6a726fdc70/samples/gles_sampler_units.cpp#L104-L108

Compare the claim at lines 37-38:

https://github.com/RandyGaul/cute_framework/blob/33d8569ea75909bb484e374e05126b6a726fdc70/samples/gles_sampler_units.cpp#L36-L39

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

pusewicz:
Fixed in d908721d432703d6be1d87b8e40725429c5f482e — added `--gles` to opt a native run into the real GLES3 backend (Linux/Windows only; confirmed it correctly fails to open a window on macOS, which has no ES-profile GL context). Without the flag the sample now says explicitly that it isn't testing the fix.

